PR #7, opened 2021-05-19
Add dockerfiles for building and translating user code.

Choc13:


docker/build/usr/bin/cc
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+
+ARTIFACTS="/usr/bin" \
+exec /lib/symbolica/cc "$@"
```
Choc13, line 5: I decided to re-introduce a wrapper script for `cc` which just set `ARTIFACTS` to `/usr/bin` as we know that's where the bits are on the docker image. That way we can keep the `cc` in `/lib/build` more flexible for those other builds.

docker/build/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    apt-get remove --purge -y software-properties-common wget && \
+    apt-get autoremove --purge -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    echo '#!/bin/bash \n$(which clang-8) "$@"' > /usr/bin/clang && \
+    chmod a+x /usr/bin/clang && \
+    echo '#!/bin/bash \n$(which llvm-link-8) "$@"' > /usr/bin/llvm-link && \
+    chmod a+x /usr/bin/llvm-link
+
+COPY ./docker/build/ .
```
Choc13, line 16: Note these copies are relative to the root of the directory, so I as I point out in the README you have to build these images in the context of the repo root too.

docker/build/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY ./lib/build/ /lib/symbolica/
```
Choc13, line 17: I decided to split the `/lib` folder in the repo root by stage, so we have a `/build` folder for all build related bits and `/translate` for the translate bits. I thought this would be a useful way to organise `lib` irrespective of Docker, but it's a problem then I can work with it in a different format too.

.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          fetch-depth: 0
+      - name: Setup dotnet SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: "5.0.202"
+      - name: Install GitVersion
+        uses: gittools/actions/gitversion/setup@v0.9.9
+        with:
+          versionSpec: "5.6.0"
```
iscottb122, line 22: I think we want version 5.6.9 due to those bugs.

lib/build/README.md
```diff
@@ -0,0 +1,21 @@
...
+From this directory run the following command:
+
+```sh
+docker build . -t symbolica/build:latest
+```
+
+## Running the image
+
+```sh
+docker run -v $(pwd)/<path-to-user-code>:/code ssvm-build:latest </path/to/user/build.sh>
```
iscottb122, line 14: This should say `symbolica/build:latest` instead of `ssvm-build:latest`. And in the example below please change it from `/tests/pass/c/average` to just `mycode` or `./mycode` or something (because the leading `/` implies root and is slighly confusing).

And maybe mention that `</path/to/user/build.sh>` is optional if you like, but I don't mind either way really.

Apart from that, LGTM, where G == Glorious.

iscottb122 (changes requested):

iscottb122 (approved):

PR #12, opened 2021-05-20
Add a macos build to the matrix for the dotnet code.

Choc13:


iscottb122:
When the next Z3 package is released, and this change can finally be merged, it should also be possible to update the reference to `Symbolica.Computation` in the Hosted repository and remove the silly shim for Z3 [here](https://github.com/SymbolicaDev/Hosted/blob/9c38f5ff14cca785039ee231a73132a822d14f37/.github/workflows/deploy.yml#L26).

iscottb122 (approved):

PR #13, opened 2021-05-24
Add translate.sh script and use it as Docker entrypoint.

Choc13:


lib/translate/translate.sh
```diff
@@ -0,0 +1,18 @@
...
+set -e
+
+outDir=./.symbolica
+mkdir -p $outDir
+
+extractFile=$(mktemp) || exit 1
+trap "rm -f $extractFile" 0 2 3 15
+
+optimizedFile=$(mktemp) || exit 1
+trap "rm -f $optimizedFile" 0 2 3 15
```
iscottb122, line 11: I wouldn't bother with this cleanup weirdness unless you really want to.

iscottb122 (approved):

lib/translate/translate.sh
Choc13: It feels like the "right thing to do", in the sense that if we want to use this script outside of docker then we should clean up temporary resources. However, I agree that the solution seems a bit inelegant somehow, but I'm not sure if that's just because I've not used it before / don't know of a "better" way.

PR #16, opened 2021-05-25
Decouple Execution from Deserialization (by moving its call to Applic…

iscottb122:
…ation) and Implementation (by flipping the dependency).


src/Deserialization/Deserializer.cs
```diff
@@ -4,19 +4,18 @@
 using System.Threading.Tasks;
 using Symbolica.Abstraction;
 using Symbolica.Deserialization.DTOs;
-using Symbolica.Execution;
 
 namespace Symbolica.Deserialization
 {
-    public sealed class Deserializer : IDeserializer
+    public static class Deserializer
```
Choc13, line 10: You're making this F# port easier by the commit. 😝

Choc13 (approved):

PR #20, opened 2021-05-28
Add editorconfig and run cleanup (which adds newlines at end of files).

iscottb122:


src/Application/Implementation/WslFileSystem.cs
```diff
@@ -43,9 +43,9 @@ private sealed class WslMountProxy : IDirectory
             public string[] GetNames()
             {
                 return DriveInfo.GetDrives()
-                    .Select(i => i.Name.Substring(0, 1).ToLowerInvariant())
+                    .Select(i => i.Name[..1].ToLowerInvariant())
```
iscottb122, line 46: To save your eyes and scroll-wheel, this is the only non-newline thing that has changed. New version of Rider has new inspections.

Choc13 (approved):

PR #24, opened 2021-06-04
Use streams instead of files for translate step.

iscottb122:


Choc13:
I think the fix for the docker build is to add `apt-get update` to the step when `clang` is installed. 

I'm confused about why the clang installation step wasn't retrieved from the cache when the preceding layers from the `llvm-base` image were, but it does somehow seem that because the contents have changed that are referenced in the `COPY` step of that intermediate builder image it has caused the `RUN` layer above it to miss the cache.

iscottb122:
Yes, that seems to have fixed it.

Choc13 (approved):

PR #26, opened 2021-06-10
Create Context in Expression instead of individual symbolic Values wh…

iscottb122:
…en returning integer.

### DIFF
--- a/src/Computation/ConstantBitVector.cs
+++ b/src/Computation/ConstantBitVector.cs
@@ -17,7 +17,10 @@ namespace Symbolica.Computation
             _value = value;
         }
 
-        public override BigInteger Integer => ToConstantUnsigned().Integer;
+        public override BigInteger GetInteger(Context context)
+        {
+            return ToConstantUnsigned().Integer;
+        }
 
         public override IProposition GetProposition(IPersistentSpace space, SymbolicBool[] constraints)
         {
--- a/src/Computation/ConstantUnsigned.cs
+++ b/src/Computation/ConstantUnsigned.cs
@@ -14,7 +14,12 @@ namespace Symbolica.Computation
             Integer = value;
         }
 
-        public override BigInteger Integer { get; }
+        public BigInteger Integer { get; }
+
+        public override BigInteger GetInteger(Context context)
+        {
+            return Integer;
+        }
 
         public override IProposition GetProposition(IPersistentSpace space, SymbolicBool[] constraints)
         {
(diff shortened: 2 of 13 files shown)


src/Computation/ConstantUnsigned.cs
```diff
@@ -14,7 +14,12 @@ private ConstantUnsigned(Bits size, BigInteger value)
             Integer = value;
         }
 
-        public override BigInteger Integer { get; }
+        public BigInteger Integer { get; }
```
Choc13, line 17: Do you still need the `Integer` property if the interface doesn't require it in this form any longer?

src/Computation/SymbolicBitVector.cs
```diff
@@ -324,15 +330,5 @@ private IValue Create(IValue other, Func<Context, BitVecExpr, BitVecExpr, BoolEx
         {
             return new SymbolicBool(c => func(c, _func(c), other.ToSymbolicBitVector()._func(c)));
         }
-
-        private BigInteger ComputeInteger()
-        {
-            using var context = new Context();
```
Choc13, line 330: Is this to alleviate the memory pressure from creating many contexts?

src/Computation/ConstantBitVector.cs
```diff
@@ -17,7 +17,10 @@ private ConstantBitVector(Bits size, IPersistentList<byte> value)
             _value = value;
         }
 
-        public override BigInteger Integer => ToConstantUnsigned().Integer;
+        public override BigInteger GetInteger(Context context)
```
Choc13, line 20: I'm guessing these are just degenerate cases that don't need a `context` to compute the integer? My F# brain is hurting because of the unused parameter. Does C# not want you to rename it to `_` to signify that it's unused?

src/Computation/ConstantUnsigned.cs
iscottb122: It was actually needed due to a small detail related to `ArithmeticRIghtShift`, however I have now added it to the `IConstantValue` interface, so it is needed for that now anyway.

src/Computation/SymbolicBitVector.cs
iscottb122: The actual alleviation of memory pressure is only really in the tests, where the `Context` is shared. The rest of this PR is really just to allow it to be passed in from outside, but that is a reasonable thing to do anyway.

src/Computation/ConstantBitVector.cs
iscottb122: C# doesn't mind due to it being on the interface. As long as at least one implementation uses the parameter you don't get a warning or anything.

Choc13 (dismissed):

Choc13 (approved):

PR #30, opened 2021-07-30
Publish snupkgs.

Choc13:


iscottb122:
My understanding is that this does nothing and is not necessary. `nuget push` automatically picks up a similarly named `snupkg` when it finds a `nupkg`. Interesting to note that my mac test setup can be flaky though.

Choc13:
So I had that inclination too. However, when I looked at the docs regarding publishing snupkgs it showed an explicit push of `.snupkg` files.

iscottb122:
Yes, I can see that the symbols packages are already being pushed. You can see the "Download symbols" link on the right of the page, for example [here](https://www.nuget.org/packages/Symbolica.Expression/0.2.0-master0025). So if anything this change is presumably going to actually fail because it will try to overwrite them.

Choc13:
Yes I just checked that too. Fine. As we were.

PR #32, opened 2021-08-05
Use wslpath utility so files do not have to be under /mnt/ directory …

iscottb122:
…to run on Windows.

### DIFF
--- a/src/Application/Implementation/WslFileSystem.cs
+++ b/src/Application/Implementation/WslFileSystem.cs
@@ -1,5 +1,4 @@
-﻿using System.IO;
-using System.Linq;
+﻿using System.Diagnostics;
 using Symbolica.Implementation.System;
 
 namespace Symbolica.Application.Implementation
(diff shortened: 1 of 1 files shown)


src/Application/Implementation/WslFileSystem.cs
```diff
@@ -20,32 +19,28 @@ public WslFileSystem(IFileSystem fileSystem)
...
+                    Arguments = $"wslpath -w {path}",
+                    CreateNoWindow = true,
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true
+                }
+            };
+
+            process.Start();
+            process.WaitForExit();
```
Choc13, line 41: Do we care that this is blocking?

Choc13 (approved):

src/Application/Implementation/WslFileSystem.cs
iscottb122: Not really.

PR #34, opened 2021-08-08
Jump straight to llvm 12.

iscottb122:


src/Representation/Functions/CountOnes.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public FunctionId Id { get; }
+        public IParameters Parameters { get; }
+
+        public void Call(IState state, ICaller caller, IArguments arguments)
+        {
+            var expression = arguments.Get(0);
+
+            var result = Enumerable.Range(0, (int) (uint) expression.Size)
```
Choc13, line 22: Do we care that the cast from `unit` to `int` will truncate large values?

Choc13 (approved):

src/Representation/Functions/CountOnes.cs
iscottb122: We don't care at the moment, so I will leave it as is for now, but you are right that it is potentially a problem.

In the future, if we want to support very large expression sizes, we will probably have to write a few specialized data structures. Z3 uses `uint` everywhere for the size of its fixed-size bit-vectors so that would be a hard upper limit, but we currently only use half of that for .Net reasons. The main ones are that `int` has to be used for array sizing and indexing, and also as the right hand side when bit-shifting left or right. So we would need to swap in some of our own implementations of these things and then we would have to make sure that things like above are made to work with the full range too. The few places where this is the case are not particularly hard to find though, so I am happy to leave them in for now.

I could be wrong, but I think it does throw a fairly obvious exception when you exceed the range, rather than just silently truncating.

iscottb122: I had a further brain background thread thought about this, and it is even less of an issue than I previous thought. It is the implementation of the `llvm.ctpop.*` (count population of ones) family of intrinsics. This means LLVM can sling in this intrinsic for any target which has some native way to count ones on a low level, probably directly in some exotic instruction we don't care about. I would imagine this probably only goes up to `llvm.ctpop.i64` or perhaps `llvm.ctpop.i128`, but I doubt we have to worry about any existing or near future processor having the hardware that would mean `llvm.ctpop.i2147483648` or above could reasonably exist. So, overflow is not something that will happen here.

PR #36, opened 2021-08-12
Map overloaded intrinsics more sensibly, and add some simple missing …

iscottb122:
…ones.

### DIFF
--- a/src/Representation/Functions/FloatRemainder.cs
+++ b/src/Representation/Functions/FloatRemainder.cs
@@ -1,0 +1,25 @@
+﻿using Symbolica.Abstraction;
+
+namespace Symbolica.Representation.Functions
+{
+    internal sealed class FloatRemainder : IFunction
+    {
+        public FloatRemainder(FunctionId id, IParameters parameters)
+        {
+            Id = id;
+            Parameters = parameters;
+        }
+
+        public FunctionId Id { get; }
+        public IParameters Parameters { get; }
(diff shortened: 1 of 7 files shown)


src/Representation/Functions/FloatRemainder.cs
```diff
@@ -0,0 +1,25 @@
...
+        public FunctionId Id { get; }
+        public IParameters Parameters { get; }
+
+        public void Call(IState state, ICaller caller, IArguments arguments)
+        {
+            var left = arguments.Get(0);
+            var right = arguments.Get(1);
+            var result = left.FloatRemainder(right);
+
+            state.Stack.SetVariable(caller.Id, result);
```
Choc13, line 22: I'm just curious right now, is `state.Stack.SetVariable(caller.Id, result)` always the same in these `Call` implementations? Don't worry I'm not going to advocate for inheritance if it is 😂

Choc13 (approved):

src/Representation/Functions/FloatRemainder.cs
iscottb122: Most of them are, but unfortunately not quite all. We could probably make `IFunction` return `IExpression` or something option-ish of this, and conditionally do the setting outside. I found the current way slightly more idiomatic for the descriptions of the semantics of the LLVM instructions, but there is really not much in it.

PR #37, opened 2021-08-16
Get struct types by name instead of just enumerating all.

iscottb122:


lib/translate/dis/Serialization/DTOs/ModuleDto.h
```diff
@@ -9,16 +9,26 @@ namespace {
...
+        StructTypeDto *threadType;
+        StructTypeDto *vaListType;
         std::vector<FunctionDto *> functions;
         std::vector<GlobalDto *> globals;
 
-        explicit ModuleDto(Module *module, DataLayout *dataLayout)
+        explicit ModuleDto(LLVMContext &context, Module *module, DataLayout *dataLayout)
                 : target(module->getTargetTriple()),
-                  pointerSize(dataLayout->getPointerSizeInBits()) {
-            for (auto &structType : module->getIdentifiedStructTypes()) {
```
iscottb122, line 19: This entire PR makes the world slightly worse, in my opinion, but it is a prerequisite to escape from the local minimum in which we currently find ourselves. A subsequent PR should remove all the c++ which is obviously a massive net win despite this ugliness. The issue is that the `getIdentifiedStructTypes` method here does not appear to have an equivalent in the `LLVMSharp` bindings, so I have to get the ones I need individually by name. Eventually I hope to get rid of all of these names entirely by improving how we get parameter and return types elsewhere, but for now this is the relatively small change that will get us where we need to be.

Choc13 (approved):

PR #38, opened 2021-08-20
Split integer and float representations so decimal and exponential no…

iscottb122:
…tations can be accepted in addition to current hexadecimal notation.


src/Representation/Operands/ConstantFloat.cs
```diff
@@ -0,0 +1,22 @@
...
+using Symbolica.Expression;
+
+namespace Symbolica.Representation.Operands
+{
+    public sealed class ConstantFloat : IOperand
+    {
+        private readonly Bits _size;
+        private readonly string _value;
+
+        public ConstantFloat(Bits size, string value)
```
iscottb122, line 11: This will be needed when we switch the deserialization over to the `LLVMSharp` bindings (in the next PR). When using the C++ bindings directly we could use `bitcastToAPInt` on `APFloat` (Arbitrary Precision) to get the bit pattern, and then we could just handle them the same as integer constants. However, the C# bindings do not seem to expose this, so we have to be able to parse any of the float [notations](https://llvm.org/docs/LangRef.html#simple-constants) it could throw at us.

src/Computation/ContextExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        private static (bool, long, ulong) Parse(bool sign, long exponent, string value)
+        {
+            var index = value.IndexOf('.');
+
+            return index == -1
+                ? (sign, exponent, ulong.Parse(value))
+                : (sign, exponent - (value.Length - 1L - index), ulong.Parse(value[..index] + value[(index + 1)..]));
+        }
+    }
+}
```
iscottb122, line 39: If you can see a better way to do this nonsense I would be keen on it. I think it is probably solid but it's not the kind of thing I like to get involved with. Note that this implementation is probably actually more powerful than strictly required, because it should handle a mix of decimal and exponential notation, although I highly doubt LLVM would ever actually generate such a thing. So perhaps this reasonable assumption would allow a simpler implementation overall or the use of some dotnet API I don't know about.

Choc13: Could another approach be to instead use the built in `Parse`/`TryParse` methods and then use `Math.Sign` etc to destructure the numeric value into the data you need?

Not sure if those LLVM string representations are parseable by `System.Double.Parse` though. 

Also, this feels like it would benefit from some tests.

iscottb122: It's not so much that the LLVM notations are not parseable. They are fairly standard so I think they probably would be. The issue is that these methods only exist in dotnet for IEEE 754 `binary32` and `binary64` (and maybe now `binary16`) but for any other sizes some precision would be lost (or incorrectly gained). The exact bit representation would not be maintained (which in general people don't care about, but in this we really do). This is why I can't even just extract the sign from the significand and then absolute value it. That extra bit might be needed.

You can see elsewhere in this PR that for 32-bit and 64-bit constant floats I do actually just parse them directly because `System.Single` and `System.Double` are compliant with IEEE 754.

But yes, I probably should write some tests.

iscottb122: Turns out I totally misunderstood what Z3 was expecting for all this, and floats are pretty damn complicated. But this should hopefully be good to go now.

src/Computation/BitsExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+            var midpoint = denominator >> 1;
+
+            var (rounded, shift) = remainder > midpoint || remainder == midpoint && !quotient.IsEven
+                ? quotient == upper - BigInteger.One
+                    ? (lower, scale + 1L)
+                    : (quotient + BigInteger.One, scale)
+                : (quotient, scale);
+
+            return context.MkFP(sign, shift, (ulong) (rounded % lower), sort);
+        }
```
iscottb122, line 75: I really, really wanted to avoid having to write anything like this, but unfortunately I don't see any way around it. To be honest, the implementation is not actually that difficult, thanks to the arbitrary precision magic of `BigInteger`, and I have written loads of tests which should cover all the branches and edge cases, but still annoying that I had to do this at all. If only I had something that could execute it symbolically for various precisions ...

src/Computation/BitsExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+            };
+        }
+
+        public static BigInteger GetNan(this Bits self, Context context)
+        {
+            var sort = self.GetSort(context);
+            var nan = ((BigInteger.One << ((int) sort.EBits + 2)) - BigInteger.One) << ((int) sort.SBits - 2);
+
+            return ConstantUnsigned.Create(self, nan).Integer;
+        }
```
iscottb122, line 29: The stuff above here is just moved from elsewhere since it made more sense to live here now that the other similar _filth_ is here.

tests/Computation.Tests/ConstantDoubleTests.cs
```diff
@@ -0,0 +1,437 @@
...
+                yield return double.MaxValue;
+                yield return double.PositiveInfinity;
+
+                foreach (var i in Enumerable.Range(-10, 10))
+                {
+                    yield return i / 3d;
+                    yield return 3d / i;
+                }
+
+                foreach (var i in Enumerable.Range(1, 10))
```
Choc13, line 269: Isn't this range already covered above?

Choc13 (approved):

src/Computation/BitsExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+        private static FPExpr CreateFloat(Context context, FPSort sort, bool sign, BigInteger significand, int exponent)
+        {
+            var (numerator, denominator) = exponent < 0
+                ? (significand, BigInteger.Pow(10, -exponent))
+                : (significand * BigInteger.Pow(10, exponent), BigInteger.One);
+
+            var precision = (int) sort.SBits;
+            var scale = precision - 1L;
+
+            var lower = BigInteger.One << (precision - 1);
```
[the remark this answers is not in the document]
iscottb122: They are really only coincidentally the same. Also I want to keep `scale` as a `long` whereas the RHS of any `<<` has to be an `int`. Also `scale` is mutated below so it seemed more sensible not to derive something else from its value on a particular line.

tests/Computation.Tests/ConstantDoubleTests.cs
Choc13: Of course. I was being stupid and read the first one as the range `[-10..10]` as you've probably guessed.

src/Computation/BitsExtensions.cs
iscottb122: For future reference, [source](https://www.exploringbinary.com/correct-decimal-to-floating-point-using-big-integers/).

PR #39, opened 2021-08-23
Switch the Deserialization over to use LLVMSharp.

iscottb122:


src/Application/Application.csproj
```diff
@@ -5,6 +5,7 @@
     <IsPackable>False</IsPackable>
     <OutputType>Exe</OutputType>
     <RootNamespace>Symbolica.Application</RootNamespace>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
```
iscottb122, line 8: I hate this and want to get rid of it. I don't think LLVMSharp is quite packaged correctly but maybe I am misunderstanding something. You know more about all this RID and dotnet build type stuff than me so we should discuss all the various bits once you are happy with everything else in this PR.

src/Application/Executor.cs
```diff
@@ -24,6 +24,7 @@ internal static class Executor
             var translateImage = Environment.GetEnvironmentVariable("SYMBOLICA_TRANSLATE_IMAGE");
 
             File.Delete(Path.Combine(directory, "symbolica.bc"));
+            File.Delete(Path.Combine(directory, ".symbolica.bc"));
```
iscottb122, line 27: Obviously the json file is gone now and replaced with a "translated" bitcode file. This seemed like a reasonable name for it, but hopefully we manage to do the "translation" internally in memory soon and it goes away again.

src/Deserialization/Deserialization.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
```
iscottb122, line 4: Has to be.

src/Deserialization/Extensions/BasicBlockRefExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            var instruction = self.FirstInstruction;
+
+            while (instruction != self.LastInstruction)
+            {
+                yield return instruction;
+                instruction = instruction.NextInstruction;
+            }
+
+            yield return instruction;
```
iscottb122, line 18: These "iterators" are really weird. They don't work like their C++ equivalents where "end" is actually "one past the end", so you can cleanly loop them. Clearly, these cannot support the case when there are zero Functions, Globals, or BasicBlocks, but I don't believe this degenerate case can ever actually happen, so simply yielding one "extra" should be fine.

src/Deserialization/Extensions/ValueRefExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Linq;
+using LLVMSharp.Interop;
+
+namespace Symbolica.Deserialization.Extensions
+{
+    internal static class ValueRefExtensions
+    {
+        public static string GetValue(this LLVMValueRef self)
+        {
+            return self.ToString().Split(' ').Last();
```
iscottb122, line 11: This is the _filth_ that I was talking about when I said I can't see a reliable way to get the "value of a value" without simply dumping and parsing it. If we could somehow find a way to do this we could probably get rid of all that floaty nonsense that went before. Might be worth a fresh set of eyes just in case.

src/Deserialization/InstructionFactory.cs
```diff
@@ -0,0 +1,221 @@
...
+        private IAttributes GetAttributes(LLVMValueRef instruction, LLVMAttributeIndex attributeIndex)
+        {
+            var attributes = instruction.GetCallSiteAttributes(attributeIndex);
+
+            var kind = _unsafeContext.GetEnumAttributeKind("signext");
+            var isSignExtended = attributes.Any(a => _unsafeContext.GetEnumAttributeKind(a) == kind);
+
+            return new Attributes(isSignExtended);
+        }
+
```
iscottb122, line 179: I want full feedback on everything in this PR with the exception of the 2 functions below. They can clearly be golfed but I want to improve their actual representations which will affect this quite a bit. I want to make those improvements in a separate PR so this change is kept inside the Deserialization space and is a fairly direct translation from the C++ equivalent.

src/Deserialization/OperandFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+            _idFactory = idFactory;
+            _unsafeContext = unsafeContext;
+        }
+
+        public IOperand Create(LLVMValueRef operand, IInstructionFactory instructionFactory)
+        {
+            return operand.Kind switch
+            {
+                LLVMValueKind.LLVMArgumentValueKind =>
+                    new Argument(int.Parse(operand.GetValue()[1..])),
```
iscottb122, line 32: Here is a use of the string dump nonsense I mention above.

src/Deserialization/OperandFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+                operand.TypeOf.GetStoreSize(_targetData).ToBits(),
+                operand.GetConstants()
+                    .Select(o => Create(o, instructionFactory))
+                    .ToArray());
+        }
+
+        private IOperand CreateConstantInteger(LLVMValueRef operand)
+        {
+            var size = operand.TypeOf.GetSize(_targetData);
+            var value = operand.GetValue();
```
iscottb122, line 120: This is the 2nd of the uses of the string nonsense, but this one is actually not too bad.

src/Deserialization/OperandFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+                    ? boolean
+                        ? BigInteger.One
+                        : BigInteger.Zero
+                    : BigInteger.Parse(value));
+        }
+
+        private IOperand CreateConstantFloat(LLVMValueRef operand)
+        {
+            var size = operand.TypeOf.GetSize(_targetData);
+            var value = operand.GetValue();
```
iscottb122, line 134: This is the final use of the string nonsense ... and you know where this goes and all the chaos that results.

src/Deserialization/UnsafeContext.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using LLVMSharp.Interop;
+
+namespace Symbolica.Deserialization
+{
+    internal sealed unsafe class UnsafeContext : IUnsafeContext
```
iscottb122, line 9: Has to be.

tests/Application.Tests/Application.Tests.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <AssemblyName>Symbolica.Application.Tests</AssemblyName>
     <RootNamespace>Symbolica.Application</RootNamespace>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
```
iscottb122, line 6: Same as the other one of these, I want to kill this. If some hack like this simply had to live in the Deserialization project, which is the immediate consumer of LLVMSharp, then I could probably live with it. But the way I have it at the moment it is the end user which has to remember to do this. Needs to be improved with a better understanding of RIDs and NuGet. I think there is some meta-packaging or something going on here that is not quite working, but I have not gotten to the bottom of it yet.

tests/Application.Tests/ExecutorTests.cs
```diff
@@ -52,7 +52,7 @@ private class TestData : TheoryData<string, bool, bool, bool>
         {
             protected TestData(string status)
             {
-                foreach (var directory in Directory.EnumerateDirectories(Path.Combine("..", "..", "..", "..", status)))
+                foreach (var directory in Directory.EnumerateDirectories(Path.Combine("..", "..", "..", "..", "..", status)))
```
iscottb122, line 55: The working path gets an extra segment when you specify the RID like above so hopefully this can be reverted too if that gets sorted.

src/Application/Executor.cs
Choc13: So just for my own sake. `symbolica.bc` is the output from the "build" stage and then `.symbolica.bc` is the translated version of this after we've run the "translate" docker image over it? I wonder if we should name it a bit differently to make the distinction clearer, even if it is only temporary.

src/Deserialization/Deserialization.csproj
```diff
@@ -10,4 +11,9 @@
     <ProjectReference Include="..\Representation\Representation.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="libLLVM" Version="12.0.0" />
+    <PackageReference Include="LLVMSharp" Version="12.0.0-beta1" />
```
Choc13, line 16: Someone likes to live dangerously. 😜

src/Deserialization/Deserialization.csproj
Choc13: 😱

src/Application/Application.csproj
Choc13: Hmm I'll have a think about this. I think I'll have to do some digging into how LLVMSharp is packaged as I don't quite get what this is doing just from looking at it. What problem did you hit before you added this?

iscottb122: The problem is the good old `Unable to load shared library 'libLLVM' or one of its dependencies.`. The shortest issue that has the workaround is [here](https://github.com/microsoft/LLVMSharp/issues/157). There is an older issue [here](https://github.com/microsoft/LLVMSharp/issues/115), including somebody who may be having my current problem of not being able to get it working on Linux (Windows is fine). There is also a similar issue for ClangSharp but with some more detail [here](https://github.com/microsoft/ClangSharp/issues/118) and this eventually leads to underlying NuGet behaviour [here](https://github.com/NuGet/Home/issues/6083).

src/Application/Executor.cs
iscottb122: We could do. It is just a string in the translate script. Here is simply where we clean it up if it exists from any previous run. I didn't put much thought into it because I am planning to kill it imminently.

src/Deserialization/Deserialization.csproj
iscottb122: We could technically minimize the code that is allowed to be unsafe by moving `UnsafeContext.cs` to its own tiny assembly with nothing else in it, since this is just a thin wrapper on `LLVMSharp.Interop` to add a few APIs for which it only provides pointer ways of interacting instead of reference ways. But it seems largely pointless to do so.

src/Deserialization/Deserialization.csproj
iscottb122: They only provide "stable" packages for libLLVM, not LLVMSharp to allow them to make whatever changes they want, so they always leave them as beta versions. I guess this is correct for strict SemVer reasons.

src/Application/Application.csproj
iscottb122: I was rubbish at searching before. The canonical issue is the one [here](https://github.com/NuGet/Home/issues/10571).

src/Deserialization/Deserializer.cs
```diff
@@ -1,26 +1,34 @@
...
+        public static IModule DeserializeModule(byte[] bytes)
         {
-            PropertyNameCaseInsensitive = true,
-            MaxDepth = int.MaxValue
-        };
+            return DeserializeModule(new UnsafeContext(), bytes);
+        }
 
-        public static async Task<IModule> DeserializeModule(Stream stream)
+        private static IModule DeserializeModule(IUnsafeContext unsafeContext, byte[] bytes)
```
Choc13, line 14: Minor, but is there much point to this `private` method, given that it seems to only be called from the `public` one above as in a one-liner. I see you're hiding away the creation of the `UnsafeContext` which seems good, but could it not be combined with the rest of this method's implementation?

src/Deserialization/IdFactory.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Symbolica.Deserialization
+{
+    internal sealed class IdFactory : IIdFactory
+    {
+        private readonly Dictionary<IntPtr, ulong> _ids;
```
Choc13, line 8: Do we care about thread safety here?

src/Deserialization/OperandFactory.cs
Choc13: If I'm not mistaken isn't `GetValue` an extension we've added? In that case, would it make sense to drop the leading `'%'` there?

src/Deserialization/OperandFactory.cs
Choc13: Ah I see in this case you don't drop the `'%'` so ignore the other comment.

src/Deserialization/UnsafeContext.cs
Choc13: Is there a more specific name for this than just `UnsafeContext`?

src/Deserialization/Deserializer.cs
iscottb122: Haha, I did this for a reason that will really annoy you. If I construct the `UnsafeContext` in the same method as the calls to `GetMemoryBuffer`, `GetTargetData`, `GetPointerSize` then these methods are never called via the interface, only directly on the known class. So then Rider tells me to just remove them from the interface, and since they don't use any instance data it then tells me to make them static methods on the class. I could do that but instead I just passed it into a method that only knows about the interface.

src/Deserialization/IdFactory.cs
iscottb122: No, deserialization of a module is single-threaded and a single `IdFactory` is used for a single deserialization.

src/Application/Application.csproj
Choc13: It looks like you've solved this for our use case by just including the specific runtime deps explicitly.

src/Deserialization/Extensions/ValueRefExtensions.cs
Choc13: I can't see anything better either unfortunately.

src/Deserialization/InstructionFactory.cs
Choc13: Ignored.

src/Deserialization/Deserializer.cs
Choc13: Lol. OOP overkill.

Choc13 (approved):

src/Deserialization/OperandFactory.cs
iscottb122: `GetValue` is used to get the full string raw "value of a value", and only in the argument case do we want to drop the first character. For getting integer and float scalars (which are the other 2 uses) we need the whole thing.

src/Deserialization/UnsafeContext.cs
iscottb122: Doesn't even have to be one thing really. Also don't actually need this class at all. Could just move the various calls to where they are used and make those methods unsafe instead. I just decided to have them all in one place but I'm not particularly tied to this design.

Choc13 (approved):

PR #46, opened 2021-08-28
Handle missing struct types and use actual type name in exception.

iscottb122:


src/Representation/Module.cs
```diff
@@ -1,50 +1,65 @@
...
+        public IStructType ThreadType => GetStructType(_threadType);
+        public IStructType VaListType => GetStructType(_vaListType);
         public IEnumerable<IFunction> Functions => _functions;
         public IEnumerable<IGlobal> Globals => _globals;
+
+        private static IStructType GetStructType((string, IStructType?) namedStructType)
+        {
+            var (name, structType) = namedStructType;
+
+            return structType ?? throw new Exception($"Struct type {name} was not found.");
```
iscottb122, line 62: While switching this exception over to the new design we discussed, I noticed that I wasn't actually properly handling the cases where the struct type is not found. So I want to get this in first and I will rebase all the other changes from here. Also note that the larger structural change here is to put the actual type name in the exception message, which is more useful for debugging.

Choc13 (approved):

PR #47, opened 2021-08-28
Move lookup of Functions and Globals onto Module.

iscottb122:


iscottb122:
This change is because it makes more sense for the coming exceptions for these lookups to be defined and thrown from the Representation side. The first 2 commits on this PR show that pretty clearly, the last 2 are just cleanup and golfing from the fact that I now pass `IModule` to some bits I didn't used to.

src/Abstraction/IModule.cs
```diff
@@ -14,7 +13,9 @@ public interface IModule
         IStructType StatType { get; }
         IStructType ThreadType { get; }
         IStructType VaListType { get; }
-        IEnumerable<IFunction> Functions { get; }
-        IEnumerable<IGlobal> Globals { get; }
+
+        IDefinition GetMain();
+        IFunction GetFunction(FunctionId functionId);
+        IGlobal GetGlobal(GlobalId globalId);
```
iscottb122, line 19: FYI I have a trivial PR lined up where I shorten the names of these various id parameters to just id (everywhere in the solution).

src/Deserialization/ModuleFactory.cs
```diff
@@ -35,10 +35,10 @@ public IModule Create(LLVMModuleRef module, Bytes pointerSize)
                 CreateStructType(module, "struct.__va_list_tag"),
                 module.GetFunctions()
                     .Select(_functionFactory.Create)
-                    .ToArray(),
+                    .ToDictionary(f => f.Id),
```
Choc13, line 38: Smallish point but I wonder if having a static `Create` function on `Module` that accepts an `IEnumerable` and effectively internalises the `ToDictionary` call would be better encapsulation here as the `Dictionary` part seems like an implementation detail of `Module`.

iscottb122: I'll do another PR for this very shortly if you don't mind. It conflicts with something else I have already done and also I will be changing this is maybe 4 places for consistency so would be better if you had a proper look at just that.

Choc13 (approved):

PR #50, opened 2021-08-29
Use custom exceptions instead of base Exception and StateException.

iscottb122:


src/Abstraction/StateException.cs
```diff
@@ -3,10 +3,10 @@
 
 namespace Symbolica.Abstraction
 {
+    [Serializable]
```
Choc13, line 6: Where is it that we need to serialise these things?

src/Computation/IrreducibleSymbolicExpressionException.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Symbolica.Computation
+{
+    [Serializable]
+    public class IrreducibleSymbolicExpressionException : Exception
```
Choc13, line 6: Do you think we should include a field that tells us the type of thing that couldn't be reduced? Could be a string here or an Enum maybe if it's a finite set?

PR #51, opened 2021-08-31
Put error enum on StateException instead of strings.

iscottb122:


src/Application/Program.cs
```diff
@@ -10,7 +10,7 @@
 if (result.IsSuccess)
     return 0;
 
-Console.WriteLine(result.Exception.Message);
+Console.WriteLine(result.Exception.Error);
```
Choc13, line 13: Reviewing this line just gave me an idea. We talked about using `ToString` still so that there's an easy way to see a string representation of the error for cases where you just want to dump the error out. Actually, I think the way to do this would be to still use the error message. For instance in `StateException` we could still call the `base` constructor with a message made up from the `StateError`. That way people who do access the `Message` on the exception will at least still get something helpful.

Choc13 (approved):

PR #52, opened 2021-08-31
Add new exception hierarchy in class libraries.

iscottb122:


src/Expression/SymbolicaRuntimeException.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace Symbolica.Expression
+{
+    [Serializable]
+    public class SymbolicaRuntimeException : SymbolicaException
```
Choc13, line 6: Is this one supposed to be used for actual exceptions? I'm not sure the name is too obvious here as all exceptions are runtime exceptions aren't they?

src/Implementation/Exceptions/UnsupportedArchitectureException.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using Symbolica.Expression;
+
+namespace Symbolica.Implementation.Exceptions
+{
+    [Serializable]
+    public class UnsupportedArchitectureException : SymbolicaException
```
Choc13, line 7: Is this not one we want to show to users?

src/Expression/SymbolicaRuntimeException.cs
iscottb122: As discussed, it's gone.

src/Implementation/Exceptions/UnsupportedArchitectureException.cs
iscottb122: Yes, and it was, but now more explicit since it is an `ErrorException`.

Choc13 (approved):

PR #53, opened 2021-09-01
Refactor DeclarationFactory and add more IoC to deserialization.

iscottb122:


iscottb122:
This is pure refactoring of some stuff that was annoying me. Should make a couple of follow-up tasks easier and a smaller diff too.

Choc13 (approved):

PR #54, opened 2021-09-01
Add printf as a no-op for now.

iscottb122:


iscottb122:
We should probably also add a few other printing functions but I've just slung this one in for now so we can improve the examples on the site a bit.

PR #55, opened 2021-09-02
Add an Implementation options record type.

iscottb122:


src/Implementation/Options.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Symbolica.Implementation
+{
+    public sealed record Options(
```
Choc13, line 3: I've called these `Parameters` elsewhere. We don't have to be consistent as such, but it might help. Which do you prefer?

iscottb122: I did look at yours, but I preferred `Options` narrowly. Can change if you prefer. `Parameters` is something else in the core Symbolica domain but it doesn't actually appear in that project so I don't mind.

Choc13: I don't have a good argument either way. If it makes more sense to call them `Options` here then I'll just update elsewhere to follow suit.

Choc13 (approved):

src/Implementation/Options.cs
iscottb122: Doesn't strictly have to match on the Hosted side if you are going to map them anyway. If `Parameters` makes more sense over there then it's probably fine to leave it. Now that I think about it a bit more I do think `Options` is slightly better here though. They are setup choices to select which implementation of things should be used, rather than runtime parameters passed to the program under test. It's marginal though.

PR #56, opened 2021-09-03
Expand the scope of the try to catch all SymbolicaException.

iscottb122:


src/Application/Executor.cs
```diff
@@ -1,95 +1,48 @@
...
-            var translateImage = Environment.GetEnvironmentVariable("SYMBOLICA_TRANSLATE_IMAGE");
-
-            File.Delete(Path.Combine(directory, "symbolica.bc"));
-            File.Delete(Path.Combine(directory, ".symbolica.bc"));
+        private readonly Options _options;
 
-            await CallExternalProcess(directory, buildImage == null
-                ? "./symbolica.sh"
-                : $"docker run -v $(pwd):/code {buildImage}");
+        public Executor(Options options)
```
Choc13, line 18: I'd have probably kept this as a `static` class with a `static` method, but I bet you knew I'd say that anyway. 😉

Choc13 (approved):

src/Application/Executor.cs
iscottb122: I was going to make a similar change on the Hosted side, since we don't really need the static entry point any more, but I am scared to now.

PR #60, opened 2021-09-06
Rename y if x else z select to differentiate from if x then y else z.

iscottb122:


src/Computation/IValue.cs
```diff
@@ -65,6 +64,7 @@ internal interface IValue
         IValue Write(ICollectionFactory collectionFactory, IValue offset, IValue value);
         IValue Xor(IValue value);
         IValue ZeroExtend(Bits size);
+        IValue IfElse(Func<Context, BoolExpr> predicate, IValue falseValue);
```
Choc13, line 67: Just curious, why don't you need to take `trueValue` as an arg here too? I'm guessing it's coming out of the `Func` somehow?

Choc13 (approved):

src/Computation/IValue.cs
iscottb122: Actually no, we are already "on" the `trueValue` here. I should probably explain all this nonsense. I suppose it is a bit odd ...

In general, implementing a fluent style is fairly natural for the operations we need to support. For example, for `left.Multiply(right)` the method is obviously called "on" `left`, which will be some implementation of `IValue`, and if necessary this will internally convert `right` into something compatible, so that it can do its thing.

However, the awkward one is `predicate.Select(trueValue, falseValue)` where things are a bit more complicated. This is called "on" the `predicate`. In the case where the `predicate` is constant it is trivial because it will simply be a `bool` and we can just return the appropriate value without even looking at what it is. When the `predicate` is symbolic the whole result is symbolic, and Z3 expresses this an an "ITE" (if-then-else). For plumbing reasons we need to convert the `trueValue` and `falseValue` to compatible things, so internally the `Select` can offload the operation to `IfElse`, so the call is "on" one of the values. We already have the predicate in `Func` form by this stage, hence the strange method signature.

This PR is just a rename for clarity because both of the previous things called `Select` were quite different, in quite a confusing way. Now, the new `IfElse` is like the python style `y if x else z`. The reason I am looking at this now is because I will have to do something slightly similar for the "power" operation I am adding.

PR #61, opened 2021-09-17
Alpine images

Choc13:
I'm not surprised the build is failing right now, because as I've currently got it the `clang-builder` image needs to be build first so that the other two images can install the various clang tools they need from it.


lib/build/clang
```diff
@@ -11,7 +11,7 @@ for x in $p
 do
     case "$x" in
         -o) set -- "$@" -o ; extension=".bc" ;;
-        -c) xlinker= ;&
+        -c) xlinker= ; set -- "$@" "$x$extension" ; extension= ;;
```
Choc13, line 14: @iscottb122 can you comment on whether or not this is correct? The reason for the change is that `;&` is a bashism and isn't POSIX compliant apparently. My understanding is that it's just doing a fall through to the default case. So I've just copied the code as a 'quick fix'. There's probably a better way though.

lib/build/clang
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
Choc13, line 1: alpine doesn't have bash

lib/build/ld
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 set -e
 
-exec ~/.symbolica/build/llvm-link ~/.symbolica/build/musl/libc.bc "$@"
+exec llvm-link-12 ~/.symbolica/build/musl/libc.bc "$@"
```
Choc13, line 4: I decided to remove the `llvm-link` indirection script, but let me know if you'd rather keep it still. I know you use these scripts to run things in other, non-docker, envs too, but this one looked like it's only uses were from other scripts in this folder.

lib/build/Dockerfile
```diff
@@ -1,19 +1,21 @@
...
+RUN apk add --no-cache git make
+
 WORKDIR $HOME/.symbolica/build
 COPY . .
 RUN ./musl.sh
 
+FROM base
+LABEL maintainer "Symbolica <dev@symbolica.dev>"
+
+# TODO: Only copy across the built musl artifacts
```
Choc13, line 17: I'll address this on a subsequent PR as I think we can save a bit more space by only copying across the compiled `libc.bc`.

lib/build/Dockerfile
```diff
@@ -1,19 +1,21 @@
...
-    apt-get remove --purge -y software-properties-common wget && \
-    apt-get autoremove --purge -y && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
+FROM alpine:3.14.2 as base
 ENV HOME /home/me
+RUN apk add --no-cache binutils
+COPY --from=symbolica/clang-builder /tmp/clang/install/bin/clang-12 /usr/local/bin/
+COPY --from=symbolica/clang-builder /tmp/clang/install/bin/llvm-link /usr/local/bin/
```
Choc13, line 5: So the approach I'm taking here is to copy just the relevant binaries for the `clang-builder` container. This means we need to make sure this one is built first so that this and translate image can consume it. My thinking here was that by combining all the clang compilation into one Docker file it would reduce duplication, both in code and in time. However, I think the time saving is minimal because it's compilation of `llvm-extract` and `clang` that take up 99% off the time here (clang being the worse offender). So I think actually, if we just have a script called `clang-cmake.sh` or something, which contains all the common bits up to the point we call `ninja` then we can inline the other bits into the `build` and `translate` images and reuse that script there as those two images don't need any of the same compiled llvm/clang binaries anyway. 

Also, once the translate stuff is sorted then we can just inline this all into the `build` image and do away with the compilation of `llvm-extract` and `opt` altogether.

Choc13: In fact, there is some time saving by combining them like this because a good chunk of the cpp.o files are used by more than one of these tools and so by the time it gets around to build `clang` it's only got 994 objects left to build, vs (I think) 2300 if it were to build clang in isolation. Still, given that these base layers shouldn't change too frequently and that translate will go soon, I'm still probably veering towards creating the shared script approach rather than the shared base image approach.

lib/build/ld
Choc13: In fact, I remember why I removed it now. It's because when I build `llvm-link` it's already named `llvm-link` and so we don't need the indirection script anymore for docker.

PR #62, opened 2021-09-17
Use sh instead of bash for all shell scripts.

Choc13:


lib/build/clang
```diff
@@ -11,7 +11,7 @@ for x in $p
 do
     case "$x" in
         -o) set -- "$@" -o ; extension=".bc" ;;
-        -c) xlinker= ;&
+        -c) set -- "$@" -c ; xlinker= ;;
```
Choc13, line 14: @iscottb122 I believe this has the same effect as the fall through behaviour did, but I've just made it more explicit as I don't believe this case should have to worry about the `extension` variable at all.

iscottb122 (approved):

PR #66, opened 2021-09-20
Make integration tests more flexible so we can check errors and symbols.

iscottb122:


iscottb122:
This is just a starting point and will hopefully hit the issue with the buffer overflow test in the docker container.

Choc13:
I think this a decent improvement. We should probably merge some of these changes into master and then we can focus on the failing buffer test.

lib/translate/translate
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env sh
 set -e
 
-optLevel=${2:-Oz}
-
 llvm-extract-12 --delete --ralias="$1" --rfunc="$1" symbolica.bc |
-opt-12 --strip-debug --loweratomic --${optLevel} --scalarizer -o=.symbolica.bc
+opt-12 --strip-debug --loweratomic "$2" --scalarizer -o=.symbolica.bc
```
Choc13, line 5: I think we should merge this change in to master first as it will have some implications on the hosted/playground side.

Choc13 (approved):

PR #69, opened 2021-09-22
Fix musl libc build

Choc13:


lib/build/Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN clangParallelJobs="${clangParallelJobs:-$(nproc)}" && \
 FROM base as musl-builder
 WORKDIR $HOME/.symbolica/build
 COPY . .
-RUN apk add --no-cache git make && \
+RUN apk add --no-cache git make findutils && \
```
Choc13, line 34: Apparently the `xargs` that comes with busybox on Alpine is a bit basic and it seems to have a smaller limit on the size of the arg that can be passed to the command. So it looked like it was doing some batching of the `llvm-link` calls and so the output we were getting was just the linking off the objects in the final batch, which is why a bunch of stuff was missing.

tests/Application.IntegrationTests/FailingTests.cs
```diff
@@ -30,15 +30,16 @@ private class TestData : TheoryData<string, string, Options, StateError, string[
...
             }
 
             private static IEnumerable<(string, string, Options, StateError, string[])> SignCases()
             {
                 return
                     from optimization in new[] {"--Oz"}
-                    from useSymbolicGarbage in new[] {false, true}
-                    from useSymbolicAddresses in new[] {false, true}
-                    from useSymbolicContinuations in new[] {false, true}
+                    from useSymbolicGarbage in new[] {true}
```
Choc13, line 40: From a bit of trial and error it appears that only have enough of a memory budget right now to run these for a single combination of the execution options. I think reducing the options permutations but having the buffer test case is better, as a temporary measure, as it will let us golf the Alpine image with a bit more confidence.

iscottb122 (approved):

PR #74, opened 2021-09-23
⛳️  Golf the LLVM CMAKE options in the Dockerfiles.

Choc13:


lib/build/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
-        -DLLVM_ENABLE_PROJECTS=clang \
         -DCMAKE_BUILD_TYPE=MinSizeRel \
+        -DLLVM_BUILD_TOOLS=OFF \
+        -DLLVM_ENABLE_PROJECTS=clang \
+        -DLLVM_INCLUDE_EXAMPLES=OFF \
+        -DLLVM_INCLUDE_TESTS=OFF \
+        -DLLVM_PARALLEL_LINK_JOBS=${llvmParallelJobs} \
         -DLLVM_TARGETS_TO_BUILD=Native \
-        -DLLVM_PARALLEL_LINK_JOBS=${clangParallelJobs} \
-        -DLLVM_ENABLE_EH=ON \
```
Choc13, line 20: According to the docs:

> Build LLVM with exception-handling support. This is necessary if you wish to link against LLVM libraries and make use of C++ exceptions in your own code that need to propagate through LLVM code. Defaults to OFF.

I don't believe we need this feature.

lib/build/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
-        -DLLVM_ENABLE_FFI=NO \
```
Choc13, line 21: This defaults to `OFF` anyway.

lib/build/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
-        -DLLVM_ENABLE_LIBCXX=NO \
```
Choc13, line 22: This defaults to `OFF` anyway.

lib/build/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
-        -DLLVM_ENABLE_PIC=YES \
```
Choc13, line 23: This defaults to `ON` anyway.

lib/build/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
-        -DLLVM_ENABLE_RTTI=YES \
```
Choc13, line 24: From the docs:

> Build LLVM with run-time type information. Defaults to OFF.

My understanding is that this is a C++ feature and I don't think we need it / use it currently.

Choc13:
This seems to have dropped 7MB off of the image size too 🥳

iscottb122 (approved):

PR #75, opened 2021-09-23
Use Release instead of MinSizeRel when building LLVM as it might actu…

Choc13:
…ally be smaller.


Choc13:
From the results of the translate image it does look like `MinSizeRel` is indeed the one we want. Switching to `Release` has increased the image size by about 35%

PR #76, opened 2021-09-23
Disable LLVM OCaml and go bindings in Docker images.

Choc13:


Choc13:
This doesn't seem to make any difference the size of the images, so I think this is probably off by default.

PR #85, opened 2021-11-08
Remove ProgramPool by making IExecutable.Run return a list of child E…

Choc13:
…xecutables.


src/Implementation/ExecutableFactory.cs
```diff
@@ -31,10 +31,7 @@ public IProgram CreateInitial(IProgramPool programPool, IModule module, Options
             var stack = new StackProxy(space, memory, CreateStack(module, options));
             var system = new SystemProxy(space, memory, CreateSystem(module));
 
-            var state = new State(programPool, module, space,
-                globals, memory, stack, system);
-
-            return new Program(() => state);
+            return new State(_ => { }, module, space, globals, memory, stack, system);
```
Choc13, line 34: Do you think we should have a `State.CreateInitial` static method?

iscottb122: Nah.

iscottb122:
I think this is good. Probably better than my design of passing `ProgramPool` down into `State` anyway. Perhaps we will just inline `StatePool` into `Executor` at a future time but leave it for now. I have some other work which makes `Executor` even simpler so maybe after that.

I hope that it is all ok with respect to stack depths but I trust that we should avoid issues as you mentioned. I'm running some big tests I have locally to try to break it, but all good so far. I became convinced for ages that there could be an issue with exiting from the multi-threaded `StatePool`, due to mismatches in signalling the `CountdownEvent`, but I think the recursive design probably is safe. Maybe we can walk through a few things together to put my mind at ease.

I can only see one likely regression in this change. Previously an exception on any 'leg' would signal an issue and all of the other 'legs' would discover this immediately on their next instruction call. They could then bail out quickly as there would be no point wasting time going any further. Now I think they only check this at the end of each `leg` when they decide whether it is worth creating any further `legs`. I'm not sure how big of an issue this is. I think we can probably work around it in the `StatePool` or a higher level either way. If all this makes your distributed work possible those are more important for us.

Finally, I pushed a branch called [remove-program-pool-notes](https://github.com/Symbolica/Symbolica/compare/remove-program-pool-notes) (which branches from this). I was going to write notes here but this was quicker and each commit on there is a note. There is nothing major, mostly renaming a few things (some back to previous names). If you are happy just update this PR by merging it back into your branch. It reduces the net changes so hopefully not too controversial.

Choc13:
Yes, you're correct about it not failing as fast now when an exception occurs, but I think we want to prioritise finding multiple issues over just the first one anyway, in which case this isn't an issue there. If we do want to still fail fast in the open source version (because it's more likely to be running on a single machine), then we can do this with a `ConcurrentBag<CancellationToken>` or something and then cancel all of them once the first exception occurs. This is more complex than the previous design for this case, but as I don't think this is the primary use case we want to support I think that's the right decision.

src/Implementation/IExecutable.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace Symbolica.Implementation
+{
+    public interface IExecutable
```
Choc13, line 5: I don't think I agree with naming this `IExecutableState`, to me it seems to fall under this category of [code smell](https://degoes.net/articles/insufficiently-polymorphic). However, I may be missing something fundamental that means that this thing really is `IExecutableState`, but from what I see in the definition of this interface, nothing tells me it should be tied to `State` and so adding that to the name seems like an abstraction leak.

iscottb122 (approved):

PR #86, opened 2021-11-09
Use command pattern

Choc13:


Choc13:
I've put this in draft mode just so you can take a look now. But I want to try and test this more in the distributed version, because I feel like 1) the tests so far haven't been too conclusive because it's only been on a single machine and 2) reworking all of the funcs in Computation might help us make a decision on whether some of the more abstract stuff I've done here is worth it, because we might find that we can reuse a lot of it there too.

PR #87, opened 2021-11-09
Bring back the integration test permutations.

iscottb122:


iscottb122:
My more exotic attempt at #78 doesn't work since Linux seems to never actually gives back the released memory. This method of pooling the `Context`s will have to do. I don't really like the fact that the integration tests could interfere with each other now (if there are any bugs in Z3) but I suppose it will have to do.

Choc13 (approved):

PR #90, opened 2021-12-07
Report the number of instructions processed.

Choc13:


src/Application/Implementation/StatePool.cs
```diff
@@ -31,27 +32,29 @@ public void Add(IExecutable executable)
...
 
-        public async Task Wait()
+        public async Task<(ulong, Exception?)> Wait()
         {
             _countdownEvent.Signal();
             await Task.Run(() => { _countdownEvent.Wait(); });
-
-            if (_exception != null)
-                ExceptionDispatchInfo.Capture(_exception).Throw();
+            return (_instructionsProcessed, _exception);
```
Choc13, line 57: The reason I've changed this to not throw now is because otherwise we won't get the instruction count. If you do still want to rethrow the exception here then we could make `InstructionsProcessed` a property on `StatePool` so that it would still be accessible even if an exception occurs. I'm in favour of this approach though because I think the interface is clearer than before and it does also have the nice bonus of getting rid of the `ExceptionDispatch.Capture` stuff.

iscottb122 (approved):

PR #91, opened 2021-12-07
Add an UnsupportedException base class.

Choc13:


src/Implementation/Exceptions/UnsupportedArchitectureException.cs
```diff
@@ -4,7 +4,7 @@
 namespace Symbolica.Implementation.Exceptions
 {
     [Serializable]
-    public class UnsupportedArchitectureException : ErrorException
+    public class UnsupportedArchitectureException : UnsupportedException
```
Choc13, line 7: This one I think is a bit of an odd one out. To me, the benefit of introducing an `UnsupportedException` type in the hierarchy is that it would easily allow us to make it possible to people to specify that they want to ignore all `UnsupportedExceptions` (as well as being able to specify specific types of unsupported operations/functions/instructions etc). However, it doesn't make sense to me to ever ignore this one because I feel like if this is the case then the user definitely wants to be told about it. 

A few alternative proposals:
1. `UnsupportedArchitectureException` inherits from `ErrorException` because actually this is a user error and they can fix it by compiling their code for a platform we do support.
1. `UnsupportedArchitectureException` inherits from `SymbolicaException` because it's actually in a singular class of its own and there's nothing the user can do to fix it. 

wdyt?

iscottb122: Hmm, I get what you are saying. It is a tricky one. I suppose if you are planning to implement some "ignore unsupported" flag then it would have to be option 2. However I'm not actually convinced that this feature is actually all that useful at all. I can see the benefit of accumulating all the errors instead of just bailing out on the first error, but I'm not sure that hiding the fact that some paths have been unexplored is actually a good thing.

iscottb122 (approved):

PR #93, opened 2021-12-08
Yield the forks after each instruction is processed on an IExecutable.

Choc13:


src/Implementation/State.cs
```diff
@@ -32,14 +32,16 @@ internal sealed class State : IState, IExecutable
             _system = system;
         }
 
-        public IEnumerable<IExecutable> Run()
+        public IEnumerable<IEnumerable<IExecutable>> Run()
         {
             _initialAction.Invoke(this);
+            yield return _forks;
```
Choc13, line 38: Let me know if this should be counted as an instruction or not.

src/Application/Implementation/StatePool.cs
```diff
@@ -30,9 +30,17 @@ public void Add(IExecutable executable)
...
-                    foreach (var child in executable.Run())
-                        if (_exception == null)
-                            Add(child);
+                    foreach (var forks in executable.Run())
+                    {
+                        if (_exception != null)
+                        {
+                            break;
+                        }
+                        foreach (var fork in forks)
```
Choc13, line 39: Writing it this way has made it a little more general purpose than we perhaps need. Specifically, this code would now support a scenario where the `State` could generate forks but that not be terminal. As in, it could fork but keep going itself. The way it is modelled is basically as a stream of forks, which  right now means a stream of empty lists until the very final one which may contain two new states if this state does indeed fork.

Choc13: In case you want to see the upshot of this in terms of them collecting stats, I've added a commit to the `executable-stats` branch so PR #90 shows what that looks like under this regime too.

src/Implementation/State.cs
iscottb122: I'm not sure that I understand the question, but if I do, I don't think it should be counted. I think this initial action have come from some instruction already and to do so here would be double counting. I may be wrong though. Would this mean you would no longer have this `yield` here?

src/Application/Implementation/StatePool.cs
```diff
@@ -30,9 +30,17 @@ public void Add(IExecutable executable)
             {
                 try
                 {
-                    foreach (var child in executable.Run())
-                        if (_exception == null)
-                            Add(child);
+                    foreach (var forks in executable.Run())
+                    {
+                        if (_exception != null)
```
iscottb122, line 35: Does this bring back the behaviour we had before #85 where an exception on any 'leg' would signal an issue and all of the other 'legs' would discover this immediately on their next instruction call. That would be good. Does this implementation do that or do we need something slightly different in the looping and breaking here?

src/Implementation/IExecutable.cs
```diff
@@ -4,6 +4,6 @@ namespace Symbolica.Implementation
 {
     public interface IExecutable
```
iscottb122, line 5: Do you think we could rename this to `IExecutableState` now and make the method on it `ExecuteNextInstruction` again? I think that is more obvious from the outside so your counting stuff isn't just guessing what this thing is doing internallay and when it is yielding.

src/Implementation/State.cs
Choc13: Yeah, if this shouldn't be counted, because it was counted by the parent state, then it means we don't need to `yield` here.

src/Application/Implementation/StatePool.cs
Choc13: Yes it does bring back that behaviour. I think that's one of the pluses for this approach, although imo it's a minor thing as there are other ways to terminate a loop anyway and it's not like this behaviour is *that* important in all scenarios.

src/Implementation/IExecutable.cs
Choc13: I still don't think so. I'll break this into two questions though as I think they have slightly different answers.

> 1. "Should we name the method `ExecuteNextInstruction`. 

This almost certainly seems wrong to me, because broadly speaking it's the same as mixing up the difference between an `Iterable` and an `Iterator`. Calling it `ExecuteNextInstruction` would signify to me that I would be expected to keep calling this until some terminating condition was met, like you would with an `Iterator`. However, what we have here is now an `Iterable` and so what the client is really doing is saying, please run yourself until you hit a terminating condition and at each stage of the process return any new child states you've created.

To highlight this with an example, I think it would be odd to write:

```csharp
foreach (var forks in executable.ExecuteNextInstruction()) 
```

> 2. "Should we rename `IExecutable` to `IExecutableState`.

In light of my answer to the above I think not and I still think generally that we shouldn't need to leak the implementation details in the interface name.

On the broader point about making it more obvious that each element of the outer `IEnumerable` is actually the result of processing one instruction, I agree with you, but I think that's actually a downside of the way we've done it here. We're kind of artificially returning an `IEnumerable` just so that the client can count the number of elements that were in it. In doing so we've made it possible for the implementation of `IExecutable.Run` to potentially produce many forks before it finally terminates, but I don't think that is behaviour that we actually want or need. 

I'm struggling to see a way to convey the semantics that each element of the `IEnumerable` represents the processing of a single instruction, without either making the method name overly verbose or just adding this to the doc string for this method. Either way, I think it's an implicit part of the interface contract that in this implementation isn't going to be enforced by the type system.

PR #94, opened 2021-12-10
Symbolize files.

Choc13:


src/Implementation/System/SymbolicDescription.cs
```diff
@@ -0,0 +1,74 @@
...
+                2U => (long)_size + offset,
+                _ => -1L
+            };
+
+            return (result, new SymbolicDescription(_size, result));
+        }
+
+        public int Read(ISpace space, IMemory memory, IExpression address, int count)
+        {
+            if (_offset < 0)
```
Choc13, line 34: I'm not sure if this is correct, but from my quick skimming of the documentation it looks like if the offset is negative then it means something's gone wrong in a previous file operation. So I introduced a new `StateError` here, which may or may not be the correct way to handle this, assuming it is even an error that should be handled here.

src/Implementation/System/SymbolicDescription.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public int GetStatus(ISpace space, IMemory memory, IStruct stat, IExpression address)
+        {
+            // TODO: Should these be symbolic values?
+            var type = Convert.ToInt32("0100000", 8);
+            var mode = Convert.ToInt32("00444", 8);
+
+            var longSize = ((Bytes)sizeof(long)).ToBits();
+            // TODO: Are these offsets still correct?
```
Choc13, line 57: @iscottb122 would appreciate your knowledge on these two TODOs

src/Implementation/System/DescriptionFactory.cs
```diff
@@ -7,16 +9,38 @@ namespace Symbolica.Implementation.System
...
+                    path = path.Add(space.CreateConstant(path.Size, (uint)Bytes.One));
+                }
+            }
+            try
+            {
+                var concretePath = new string(ReadCharacters().ToArray());
+                return CreateFile(concretePath)
+                    ?? CreateDirectory(concretePath);
+            }
+            catch
```
Choc13, line 40: I'll change this to catch the irreducible exception, even though this isn't the way we want to handle it. Might as well at least make my dirty hacks correct for now.

PR #96, opened 2021-12-20
Simplify value interfaces which removes stored lambdas from Computation.

iscottb122:


src/Computation/IValue.cs
```diff
@@ -1,21 +1,14 @@
...
-        IBitwise AsBitwise();
-        IBitVector AsBitVector(ICollectionFactory collectionFactory);
-        IUnsigned AsUnsigned();
-        ISigned AsSigned();
-        IBool AsBool();
-        IFloat AsFloat();
-        IValue IfElse(IBool predicate, IValue falseValue);
+        BitVecExpr AsBitVector(Context context);
+        BoolExpr AsBool(Context context);
+        FPExpr AsFloat(Context context);
```
iscottb122, line 12: A lot of boilerplate in this PR, but it is the complexity-killing kind we can accept, and the win is that this interface becomes much simpler, then the removal of stored lambdas comes for free from that.

src/Computation/Expression.cs
```diff
@@ -2,308 +2,500 @@
...
+                    32U => _value,
+                    64U => new ConstantDouble(v),
+                    _ => new FloatConvert(size, _value)
+                },
+                v => (uint) size switch
+                {
+                    32U => new ConstantSingle((float) v),
+                    64U => _value,
+                    _ => new FloatConvert(size, _value)
+                });
```
iscottb122, line 112: This float thing is the one thing that I would say becomes a bit worse, but I think we just have to eat this.

src/Computation/Expression.cs
```diff
@@ -2,308 +2,500 @@
...
 using Symbolica.Expression;
 
 namespace Symbolica.Computation
 {
     internal sealed class Expression : IExpression
     {
         private readonly ICollectionFactory _collectionFactory;
-        private readonly Lazy<BigInteger> _constant;
-        private readonly IBool[]? _constraints;
+        private readonly IValue[]? _constraints;
```
Choc13, line 18: An optional array seems like it introduces an unnecessary degree of freedom to me, as an empty array already indicates the absence of any values. I checked the code out and made this non-nullable. It seems like it still works with this change, but obviously forces a few changes in places that call the constructor etc because they obviously now need to supply `Array.Empty<>`. Personally, with my F# hat on, I prefer to eliminate the nulls and be explicit with `Array.Empty` at the call sites than save on a few characters of code. It's possible I've overlooked something more fundamental behind the reason for making these values nullable here though.

src/Computation/Expression.cs
```diff
@@ -316,36 +508,82 @@ private BigInteger AsConstant()
...
-                func(_value, y._value, z._value), Concat(_constraints, Concat(y._constraints, z._constraints)));
+            return _value is IConstantValue cx && y._value is IConstantValue cy && z._value is IConstantValue cz
+                ? new Expression(_contextFactory, _collectionFactory,
+                    constant(cx, cy, cz), null)
+                : new Expression(_contextFactory, _collectionFactory,
+                    symbolic(_value, y._value, z._value), Concat(_constraints, Concat(y._constraints, z._constraints)));
         }
 
-        private static IBool[]? Concat(IBool[]? left, IBool[]? right)
+        private static IValue[]? Concat(IValue[]? left, IValue[]? right)
```
Choc13, line 586: Also, if these become non-nullable then the implementation of this method is simplified.

src/Computation/Expression.cs
```diff
@@ -2,308 +2,500 @@
...
+        }
+
+        private static IValue GetValue(IPersistentSpace space, IValue value, IValue[] constraints)
+        {
+            using var model = space.GetModel(constraints);
+
+            return ConstantUnsigned.Create(value.Size, model.Evaluate(value));
         }
 
         public static IExpression Create(IContextFactory contextFactory, ICollectionFactory collectionFactory,
```
Choc13, line 501: Similarly, I noticed that `constraints` here could also be made non-nullable given an empty enumerable already indicates no values and also in this case it looks like there is currently the possibility of dereferencing a `null` value at `constraints.Select`.

src/Computation/Expression.cs
```diff
@@ -2,308 +2,500 @@
...
-            return Create(c => c.GetValue((IPersistentSpace) space, _constraints ?? Array.Empty<IBool>()));
+            return Unary(
+                v => GetValue((IPersistentSpace) space, v, _constraints ?? Array.Empty<IValue>()),
+                v => v);
         }
 
         public IProposition GetProposition(ISpace space)
         {
-            return _value.AsBool().GetProposition((IPersistentSpace) space, _constraints ?? Array.Empty<IBool>());
+            return _value is IConstantValue c
```
Choc13, line 46: This type switching in here, to decide if something is constant or symbolic, made me wonder if we should have two implementations of `IExpression`, a `ConstantExpression` and a `SymbolicExpression`. You've probably already thought about this though, so just curious why you chose this design?

src/Computation/Values/BitVector.cs
```diff
@@ -0,0 +1,18 @@
+using Microsoft.Z3;
+using Symbolica.Expression;
+
+namespace Symbolica.Computation.Values
+{
+    internal abstract class BitVector : Integer
```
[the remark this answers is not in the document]
iscottb122: It's one of the few times when I prefer inheritance. It's the simplest way I could find to have small and DRY "command-pattern" implementations of `IValue`.

src/Computation/Expression.cs
iscottb122: The nullable is gone now.

src/Computation/Expression.cs
iscottb122: This is gone now. I thought there were performance implications but it was something unrelated.

src/Computation/Expression.cs
iscottb122: I have switched over to this design now. I had something similar to this way back but other issues caused a problem with it. These are no longer valid though so I think it is better.

src/Computation/SymbolicExpression.cs
```diff
@@ -0,0 +1,391 @@
...
+
+        private static BigInteger AsConstant(Expr expr)
+        {
+            return expr.IsNumeral
+                ? ((BitVecNum) expr).BigInteger
+                : throw new IrreducibleSymbolicExpressionException();
+        }
+
+        public static SymbolicExpression Create(IContextFactory contextFactory, ICollectionFactory collectionFactory,
+            IValue value, IEnumerable<Func<IExpression, IExpression>>? constraints)
```
Choc13, line 378: I'd personally be tempted to make `constraints` non-nullable, given that (and I think this applies here) an empty enumerable serves the same purpose.

Choc13 (dismissed):

src/Computation/SymbolicExpression.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        public static SymbolicExpression Create(IContextFactory contextFactory, ICollectionFactory collectionFactory,
+            IValue value, IEnumerable<Func<IExpression, IExpression>> constraints)
+        {
+            var unconstrained = new SymbolicExpression(contextFactory, collectionFactory,
+                value, Array.Empty<IValue>());
+
+            return new SymbolicExpression(contextFactory, collectionFactory,
+                value, constraints
+                    .Select(c => ((SymbolicExpression) c(unconstrained)).Value)
```
Choc13, line 385: I meant to say previously. How do we know that this cast is always safe?

iscottb122: That invariant is enforced within this class. If the input to the `Func` is a `SymbolicExpression` then the output is guaranteed to be too. However, I will actually change the cast to `IValueExpression` since that works just as well.

Choc13 (approved):

PR #99, opened 2022-01-06
Create a base C++ build image.

Choc13:


lib/build/cpp/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        -DLIBCXX_HAS_GCC_S_LIB=OFF \
+        -DLIBCXX_HAS_MUSL_LIBC=ON \
+        -DLIBUNWIND_ENABLE_SHARED=OFF \
+        -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind" && \
+    CODEGEN=1 ninja -C build cxx cxxabi unwind && \
+    # # TODO: Enable testing once it's working
+    # # ninja -C build check-cxx check-cxxabi check-unwind && \
+    ninja -C build install-cxx install-cxxabi install-unwind && \
+    mkdir -p ~/.symbolica/build/c++ && \
+    find build/libunwind -name "*.o.bc" -print0 | xargs -0 ~/.symbolica/build/llvm-link -o ~/.symbolica/build/c++/libunwind.bc && \
```
Choc13, line 36: @iscottb122 I've literally just slung these llvm link lines in. They're currently not working for me, but I thought I'd put them here in case you want to take over from this point so you can make some progress based off of the rest of it.

Specifically, when I commented out the lines after `ninja -C build install` and then tried to `llvm-link` libunwind, I got the following error:

```sh
llvm-link-12: build/libunwind/src/CMakeFiles/unwind_static.dir/UnwindRegistersRestore.S.o.bc:1:1: error: expected top-level entity
# 1 "/tmp/llvm-project/libunwind/src/UnwindRegistersRestore.S"
^
llvm-link-12: error:  loading file 'build/libunwind/src/CMakeFiles/unwind_static.dir/UnwindRegistersRestore.S.o.bc'
```

I've literally just done this so apologies for not taking any time to investigate this one of my own.

lib/build/cpp/c++
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Choc13, line 1: I'm not too tied to having this as a separate file, I mostly did it this way for now just to keep things separate whilst we're still figuring bits out. It does possibly seem slightly nicer (and kind of inline with usual linux conventions from what I can tell), to have a separate `c++` wrapper script.

lib/build/cpp/Dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG BUILD_BASE_IMG_TAG
```
Choc13, line 1: To build this, run `docker build . -f Dockerfile.cpp --build-arg BUILD_BASE_IMG_TAG=0.2.0-master0076 -t symbolica/build-cpp:dev`

lib/build/cpp/cxx
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+set -e
+
+exec ~/.symbolica/build/cc -nostdinc++ "$@"
```
Choc13, line 4: Do you think we should ultimately be calling through to `clang++` here instead of `clang-12` because from what I've read `clang++` will set a few other C++ specific flags such as `-lc++` when linking.

lib/build/cpp/cxx
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+set -e
+
+# There is something more mysterious going on here wrt crt* objects
+# The compiler seems to be finding the ones in sysroot/lib and adding them
+# to the linker. So will it also find crtbegin from compiler-rt if we tell it to
+# look there, or would we need to rebuild clang with USE_COMPILER_RT enabled?
```
Choc13, line 7: So I think we can probably do something better re these crt{begin/end} files. For starters we could just relocate the compiler-rt ones and overwrite the dummy placeholders we already have. Or possibly we could rebuild clang with `CLANG_DEFAULT_RTLIB=compiler-rt` which I believe configures it use compiler-rt generated bits instead. 

On a related note there is also `CLANG_DEFAULT_CXX_STDLIB` which if we set to `libc++` I think would then bake that flag in to the compiler. These two points, plus a few other things make me wonder whether we should do a fresh build of clang once we've bootstrapped our own sysroot with our musl libc and built libc++, compiler-rt etc so that we then have a version of clang that is more aware of our custom built environment and it would also be more consistent across environments. Probably one to ponder for a bit and to implement another day when we've got some stability around what we exactly need in the toolchain for c++.

lib/build/cpp/cxx-bootstrap
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Choc13, line 1: @iscottb122 I don't particularly love this approach, but it works and given that you're doing some reconfiguring of the sysroot at the musl level then it's possible that when we merge those changes in there will be a better way to handle this situation that's not available to me right now, so I've not put much effort into making this any nicer currently.

lib/build/clang
```diff
@@ -4,22 +4,25 @@ set -e
 p="$@"
 set --
 
+CLANG=${CLANG:='clang-12'}
+crtBeginXlinker="$CRT_BEGIN_XLINKER"
+crtEndXlinker="$CRT_END_XLINKER"
```
Choc13, line 9: Don't love this either, see below for a longer discussion on how we can hopefully get `crt{begin/end}` working without us having to explicitly specify anything in our wrapper scripts.

PR #103, opened 2022-01-12
Try using musl-clang instead of the manual include and linker nonsense.

iscottb122:


lib/build/cc
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 set -e
 
-XLINKER="$XLINKER -Xlinker -lc -Xlinker -nostdlib -Xlinker -static" \
-exec ~/.symbolica/build/clang -nostdinc -nodefaultlibs --sysroot ~/.symbolica/build/sysroot -isystem ~/.symbolica/build/sysroot/include -O0 -Xclang -disable-O0-optnone -fPIC "$@"
+exec ~/.symbolica/build/sysroot/bin/musl-clang -O0 -Xclang -disable-O0-optnone -fPIC -static "$@"
```
Choc13, line 4: This seems like it's going to bypass the clang wrapper script we have now. Shouldn't it be in the wrapper script that we call `musl-clang`?

Choc13: Ah no, because musl was built using the clang wrapper, then it has itself wrapped that script thinking that is the system compiler. I see now.

PR #104, opened 2022-01-18
Rejig the structure of lib to support languages.

iscottb122:


.gitignore
```diff
@@ -2,5 +2,7 @@
 *.user
 .idea/
 .ionide/
-bin/
-obj/
+/src/*/bin/
```
Choc13, line 5: Do you think it would be better to have an opt-in pattern instead for the `bin` dirs inside `lib/*`?

lib/c/build/Dockerfile
```diff
@@ -26,16 +26,15 @@ RUN llvmParallelJobs="${llvmParallelJobs:-$(nproc)}" && \
...
 # and it ends up batching the calls to llvm-link which results in us only producing
 # a libc.bc file with the contents of the final invocation in it.
-RUN apk add --no-cache --virtual musl-build-deps git make findutils && \
-    cd $HOME/.symbolica/build/install &&  \
-    ./musl.sh && \
+RUN apk add --no-cache --virtual musl-build-deps findutils git make && \
+    cd $HOME/.symbolica && \
+    ./install-musl.sh && \
     cd / && \
-    rm -rf $HOME/.symbolica/build/install && \
```
Choc13, line 38: Are we not cleaning up installation scripts anymore?

lib/c/build/bin/cc-link
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Choc13, line 1: I was wondering what you were going to call this file. In my mind it's `llvm-ld`, but you've probably thought more about this now than I have.

lib/cpython/build/bin/cpython-cc
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+set -e
+
+SYSROOT=$(eval echo ~/.symbolica)
+
+exec ~/.symbolica/bin/cc -I$SYSROOT/include/python3.8 -DCPYTHON_HOME=L\"$SYSROOT\" -DCPYTHON_EXEC=L\"$SYSROOT/bin/python3\" -DCPYTHON_PATH=L\"$SYSROOT/lib/python3.8\" "$@" -lpython3.8
```
Choc13, line 6: "${SYSROOT}/bin/cc"?

Choc13: I've been wondering more generally whether we expect the user to set $SYMBOLICA_HOME in their environment and then use that throughout these scripts. Certainly makes things easier in Docker land where we can set it inside the container when we build it. I think other advantages would be that it would potentially allow side-by-side installs of different versions on a system.

lib/c/build/install-musl.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
```
Choc13, line 1: Probably worth setting `SYSROOT` (or `SYMBOLICA_HOME`) in this script and using that throughout (or pending my other comment, relying on it already being set in the current environment.

.gitignore
iscottb122: By doing something fancy here? Or by sticking another gitignore inside `lib`?

lib/c/build/bin/cc-link
iscottb122: To be honest, my naming of all these things is actually not exactly consistent. I couldn't quite figure out how to make it so but I will have another quick look and see if your suggestion helps.

.gitignore
Choc13: By doing something here like

```.gitignore
bin/
obj/
!/lib*/bin
```

lib/c/build/Dockerfile
iscottb122: I brought it back.

lib/c/build/bin/cc-link
iscottb122: I chose `libc-link` which finally makes things consistent at least. I think my original choice of using `ld` was also a bit silly and misleading so I have killed that pattern.

lib/c/build/install-musl.sh
iscottb122: I'd rather leave this env var stuff alone for now, but maybe soon as a separate PR. I did have something like this in the past and it was pretty annoying.

Choc13 (approved):

PR #105, opened 2022-01-19
Add some C++ support.

iscottb122:


lib/cpp/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM symbolica/build:${BUILD_BASE_IMG_TAG}
+LABEL maintainer "Symbolica <dev@symbolica.dev>"
+
+COPY . $HOME/.symbolica
+
+RUN apk add --no-cache --virtual libcxx-build-deps cmake git ninja python3 && \
+    cd $HOME/.symbolica && \
+    ./install-libcxx.sh && \
+    rm -rf llvm-project install-libcxx.sh && \
+    cd / && \
```
Choc13, line 11: This line is actually redundant, but fine.

lib/cpp/build/bin/clang++
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Choc13, line 1: Let's remove this indirection and just inline it in the two places it's used to keep it simple.

lib/cpp/build/install-libcxx.sh
```diff
@@ -0,0 +1,36 @@
...
+    -DLIBCXX_HAS_GCC_S_LIB=OFF \
+    -DLIBCXX_HAS_MUSL_LIBC=ON \
+    -DLIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=ON \
+    -DLIBCXXABI_ENABLE_EXCEPTIONS=OFF \
+    -DLIBCXXABI_ENABLE_SHARED=OFF \
+    -DLIBCXXABI_ENABLE_STATIC_UNWINDER=ON \
+    -DLIBCXXABI_USE_LLVM_UNWINDER=ON \
+    -DLIBUNWIND_ENABLE_SHARED=OFF \
+    -DLIBUNWIND_ENABLE_THREADS=OFF \
+    -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi;libunwind"
```
iscottb122, line 30: I left out `-DLLVM_ENABLE_LIBCXX=ON` here but we may want to add it back in so we don't surprise ourselves in future if we ever want to use our build `clang++` directly.

lib/cpp/build/install-libcxx.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env sh
+set -e
+
+git clone --depth 1 --branch llvmorg-12.0.1 https://github.com/llvm/llvm-project.git
+
+cd llvm-project
+mkdir build
+CODEGEN=1 NO_LLVM=1 CC=~/.symbolica/bin/cc CXX=~/.symbolica/bin/clang++ \
```
Choc13, line 8: I think we might want `CXXFLAGS="-fno-threadsafe-statics"` here so that the bitcode doesn't generate `__cxa_guard_acquire` and friends.

Choc13: Well let's test it first and see if we actually do hit it in any user code.

lib/cpp/build/Dockerfile
Choc13: And let's add `rm -rf /tmp/*` to make sure there's no detritus left in there.

lib/translate/Dockerfile
```diff
@@ -5,24 +5,22 @@ LABEL maintainer "Symbolica <dev@symbolica.dev>"
...
-        -DLLVM_TARGETS_TO_BUILD=Native \
-        /tmp/llvm-project/llvm && \
-    ninja -j ${llvmParallelJobs} install-opt install-llvm-extract && \
-    mv /usr/local/bin/opt /usr/local/bin/opt-12 && \
+        -DLLVM_TARGETS_TO_BUILD=Native && \
+    ninja -C llvm-project/build -j ${llvmParallelJobs} install-llvm-extract install-opt && \
     mv /usr/local/bin/llvm-extract /usr/local/bin/llvm-extract-12 && \
-    cd / && \
-    rm -rf /tmp && \
+    mv /usr/local/bin/opt /usr/local/bin/opt-12 && \
```
Choc13, line 22: How has this been working up until now if the name should have been `opt-12`?

Choc13: Oh you rearranged the lines. Ignore.

Choc13 (approved):

PR #112, opened 2022-01-20
Make install scripts more amenable to future dockerization.

iscottb122:
I see from your docker work that you like to do things like these in separate layers, so I have split the scripts to allow that (if we ever get round to it). It's probably nicer this way for working locally anyway.

### DIFF
--- a/lib/ecl/build/install-ecl.sh
+++ b/lib/ecl/build/install-ecl.sh
@@ -1,15 +1,7 @@
 #!/usr/bin/env sh
 set -e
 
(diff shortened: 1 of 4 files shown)


lib/ecl/build/install-ecl.sh
```diff
@@ -1,16 +1,8 @@
 #!/usr/bin/env sh
 set -e
 
-wget -qO- https://gmplib.org/download/gmp/gmp-6.1.2.tar.xz | tar -xJf - --one-top-level=gmp --strip-components=1
```
iscottb122, line 4: Also, `--one-top-level=` is bash-only, so I changed it to a portable way. Everything else in this PR is pure copy-paste.

Choc13 (approved):

PR #118, opened 2022-01-24
Add some C# editorconfig rules to make IDEs more consistent.

Choc13:


iscottb122 (approved):

.editorconfig
```diff
@@ -13,3 +13,10 @@ trim_trailing_whitespace = true
 [*.{build,config,csproj,dbml,json,nuspec,proj,props,StyleCop,targets,xml,yml}]
 indent_size = 2
 tab_width = 2
+
+# .NET formatting rules
+[*.{cs,vb}]
+dotnet_sort_system_directives_first = true
+dotnet_separate_import_directive_groups = false
+dotnet_style_namespace_match_folder = true
+csharp_space_after_cast = true
```
Choc13, line 22: There are a lot more style settings we can set, as documented [here](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#spacing-options), but I've just added the minimal set now to get my IDE to be be consistent with yours based on the files I've recently edited.

PR #119, opened 2022-01-24
Try adding fast path for reading written value.

iscottb122:


src/Computation/Reader.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Symbolica.Computation
+{
+    internal sealed class Reader : IReader
+    {
+        private readonly ISymbolicExpression _buffer;
+        private readonly IReader _defaultReader;
+        private readonly IExpression _mask;
+        private readonly IExpression _value;
+
+        public Reader(IReader defaultReader, ISymbolicExpression buffer, IExpression mask, IExpression value)
```
Choc13, line 12: `ImmediateReadAfterWriteReader`? 🤣

iscottb122: `WriteReader`

Choc13 (approved):

PR #122, opened 2022-01-26
Golf the new write optimization a bit using a little inheritance.

iscottb122:


iscottb122:
I think we should probably adopt a slightly different naming strategy here. Something like changing `SymbolicExpression` to just `Expression`, `SymbolicWriteExpression` to just `WriteExpression`, the new `SymbolicArbitraryExpression` to just `SymbolicExpression`. Basically it is symbolic by default unless there is some mad optimization going on.

Choc13:
I'm afraid (and probably unsurprisingly) I'm not sure about this one. I think it feels like inheritance purely for the sake of code reuse (well not even that really because there was no duplication before, but I guess a lot of "boilerplate" because most methods on `SymbolicWriteExpression` were calling `.AsSymbolic().Foo()`). 

I think the fact that the names are obvious is a sign that the inheritance hierarchy is artificial too. Another sign for me is that we've only made `Read` and `Write` abstract because they happen to be the ones that vary between our current set of implementations. I think if this was a more universal abstraction then more of them would naturally be `abstract` too. 

Overall I think I prefer the "boilerplate" here because the code is less obscure even if it is more verbose.

iscottb122:
It's actually only `Read`, because even `Write` is actually the same, but I will have another look.

PR #130, opened 2022-01-29
Upgrade everything including dotnet, C#, Z3.

iscottb122:


.github/workflows/build-and-publish.yml
```diff
@@ -17,13 +17,17 @@ jobs:
       uses: actions/checkout@master
       with:
         fetch-depth: 0
+    - name: Setup dotnet SDK
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: "6.0.101"
```
iscottb122, line 23: The inclusion of the `global.json` means that I had to add this here, otherwise `GitVersion` complains that it can't find a permitted dotnet to do its thing. Once they add dotnet 6 to all the virtual environments we use, we could actually get rid of this step from everywhere. I didn't realise that dotnet is installed by default.

Choc13: I actually think this is a good thing and it means global.json is doing its job. From a purity point of view, I'd rather they didn't preinstall anything (but maybe if I had to wait for everything to be installed on each build I'd change my mind). But things like the dotnet version I think it's better if we're explicit about the version that we want installed.

Choc13 (approved):

Choc13 (approved):

PR #132, opened 2022-01-31
Encapsulate the ConstantUnsigned and ConstantSigned logic.

iscottb122:


src/Computation/Values/Constants/ConstantSigned.cs
```diff
@@ -50,11 +50,56 @@ public ConstantDouble AsDouble()
...
+    }
+
+    public ConstantSigned Remainder(ConstantSigned value)
+    {
+        return Create(Size, _value % value._value);
+    }
+
+    public ConstantSigned ShiftRight(ConstantUnsigned value)
+    {
+        return Create(Size, _value >> (int) (BigInteger) value);
```
Choc13, line 90: How does `BigInteger` determine when to do an arithmetic vs logical right shift?

Choc13 (approved):

src/Computation/Values/Constants/ConstantSigned.cs
iscottb122: They are infinite/arbitrary width so there is no single 'high bit' to consider as the sign. It just works because in this class the value is interpreted as a signed integer, whereas in the other one it is interpreted as unsigned. So the result is always normalized to the appropriate value with respect to the size (which has to be stored separately alongside the raw `BigInteger`).

PR #133, opened 2022-01-31
Move efficient write-read logic from Expression to Value layer.

iscottb122:


Choc13 (approved):

src/Computation/Values/Symbolics/Write.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        return Or(And(_writeBuffer, Not(writeMask)), writeData);
+    }
+
+    private IValue Mask(IValue offset, Bits size)
+    {
+        return ShiftLeft(ZeroExtend(Size, Not(ConstantUnsigned.Create(size, BigInteger.Zero))), offset);
+    }
+
+    private static IValue And(IValue left, IValue right)
```
Choc13, line 54: These functions seem generally useful, should we move them to a static class (I think you lot call it a factory 😉) so they can be used throughout these Value classes? The other option would be to have static factory methods on `And` etc and hide the constructors, but then having them in the `Symbolics` namespace would be possibly incorrect.

Choc13 (approved):

src/Computation/Values/Symbolics/Write.cs
iscottb122: I was planning to maybe do the latter of these options, but I want to see what it all looks like once the AND\OR changes are in and the two `IExpression` implementations have hopefully been rolled into one. So we'll leave it for now since they are only used here.

Choc13 (approved):

PR #134, opened 2022-02-02
Make proper value implementations which handle logic internally.

iscottb122:


src/Computation/Values/Symbolics/Write.cs
```diff
@@ -55,60 +55,4 @@ private IValue Mask(IValue offset, Bits size)
...
-            ? l.AsUnsigned().Xor(r.AsUnsigned())
-            : new Xor(left, right);
-    }
-
-    private static IValue ZeroExtend(Bits size, IValue value)
-    {
-        return value is IConstantValue v
-            ? v.AsUnsigned().Extend(size)
-            : new ZeroExtend(size, value);
-    }
```
iscottb122, line 113: Goodbye.

src/Computation/Values/Value.cs
```diff
@@ -0,0 +1,49 @@
...
+    public abstract FPExpr AsFloat(Context context);
+
+    public static IValue Select(IValue predicate, IValue trueValue, IValue falseValue)
+    {
+        return Ternary(predicate, trueValue, falseValue,
+            (p, t, f) => p.AsBool() ? t : f,
+            (p, t, f) => new Select(p, t, f));
+    }
+
+    protected static IValue Unary(IValue x,
```
Choc13, line 23: `protected static` 🤨

Choc13 (approved):

PR #139, opened 2022-02-08
Actually use the value polymorphism to avoid float type matching.

iscottb122:


src/Computation/BitVecExprExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static BigInteger AsConstant(this BitVecExpr self)
+    {
+        var expr = self.Simplify();
+
+        return expr.IsNumeral
+            ? ((BitVecNum) expr).BigInteger
+            : throw new IrreducibleSymbolicExpressionException();
+    }
+}
```
iscottb122, line 17: I could just make this a `public static` on `Integer` instead. Couldn't decide.

Choc13 (approved):

PR #140, opened 2022-02-08
Fix bug where AND and XOR are not treated as bitwise.

iscottb122:


src/Computation/BitVecExprExtensions.cs
```diff
@@ -14,4 +14,9 @@ public static BigInteger AsConstant(this BitVecExpr self)
             ? ((BitVecNum) expr).BigInteger
             : throw new IrreducibleSymbolicExpressionException();
     }
+
+    public static BoolExpr AsBool(this BitVecExpr self, Context context)
+    {
+        return context.MkNot(context.MkEq(self, context.MkBV(0U, self.SortSize)));
+    }
 }
```
iscottb122, line 22: As mentioned, I quite like the resulting syntax, but neither of these really need to be extension methods.

src/Computation/Values/And.cs
```diff
@@ -22,7 +22,9 @@ public override BitVecExpr AsBitVector(Context context)
 
     public override BoolExpr AsBool(Context context)
     {
-        return context.MkAnd(_left.AsBool(context), _right.AsBool(context));
+        return _left is Bool || _right is Bool
+            ? context.MkAnd(_left.AsBool(context), _right.AsBool(context))
```
Choc13, line 26: Does this to unnecessary constraints being generated for the one that is already a `Bool`?

tests/Computation.Tests/NormalFloatTests.cs
```diff
@@ -4,11 +4,11 @@
 using Symbolica.Expression;
 using Xunit;
 
-namespace Symbolica.Computation.Values;
+namespace Symbolica.Computation;
 
 public class NormalFloatTests
 {
-    private static readonly IContextFactory ContextFactory = new SharedContextFactory();
```
Choc13, line 11: How come we don't need to use `SharedContextFactory` here anymore?

iscottb122: We never really needed it here since the 10s of `Context`s used by these particular tests would have been fine anyway. The reason we needed it to exist at all was that the 1000s of `Context`s needed for all the other tests caused memory issues. So it was only really used here because they were all the same pattern.

I can get rid of this thing entirely (within this test project) now because I have moved the tests from the wrapper `Expression` layer to each `IValue` implementation instead. This means we can control the instantiation of `Context` ourselves, and just use a single static one per test class. This still results in 10s of `Context`s in total, which is fine from a memory point of view.

Unfortunately we do still need the pattern of allowing injection of the `IContextFactory` implementations, rather than just allowing `Computation` to construct `Context`s whenever it needs one, because the `Application` integration tests still has memory issues (since the native dependency is not unloaded properly between runs). Annoying.

src/Computation/Values/And.cs
Choc13: Yeah, I was meaning that if one side is already `Bool` then can we avoid calling `AsBool` on it, but I see that as `AsBool` converts to `BoolExpr` so it's needed regardless.

Choc13 (approved):

Choc13 (approved):

PR #141, opened 2022-02-10
Create an IContext interface to encapsulate any additional behaviour.

iscottb122:


src/Application/Computation/Model.cs
```diff
@@ -6,26 +6,27 @@
 
 namespace Symbolica.Computation;
 
-internal sealed class Model : IModel
+internal sealed class Model<TContext> : IModel
```
Choc13, line 9: What's the generic needed for?

iscottb122: Oops. I moved it now. I think the rest are where they need to be.

Choc13 (approved):

PR #143, opened 2022-02-11
Segregate creation methods on context by type constraint.

iscottb122:
This should hopefully be the last of these large brainless PRs before the actual improvements can come.


src/Computation/IContext.cs
```diff
@@ -5,6 +5,11 @@ namespace Symbolica.Computation;
 
 public interface IContext : IDisposable
 {
-    TResult Execute<TResult>(Func<Context, TResult> func)
-        where TResult : Z3Object;
+    Solver CreateSolver(Func<Context, Solver> func);
```
iscottb122, line 8: The whole point of this PR is that I want to control creation of the solver within the context.

tests/Application.IntegrationTests/PooledContext.cs
```diff
@@ -22,9 +22,20 @@ public void Dispose()
...
+    public TSort CreateSort<TSort>(Func<Context, TSort> func)
+        where TSort : Sort
+    {
+        return _context.CreateSort(func);
+    }
+
+    public TExpr CreateExpr<TExpr>(Func<Context, TExpr> func)
+        where TExpr : Expr
+    {
+        return _context.CreateExpr(func);
```
iscottb122, line 39: It's pretty clear to me now that I should not have killed `IContextHandle` from before, and should instead allow that to vary separately from the new `IContext`, which can just have one by composition. I thought I could do everything on one interface but it is actually stupid because I will have to wrap all the methods like this, and it also means I would have to unnecessarily make things public that otherwise don't have to be on my following work. If everything else is ok on this PR I will do another small follow-up with that change.

iscottb122: Never mind, just done it on this PR otherwise other things don't really make much sense.

Choc13 (approved):

PR #144, opened 2022-02-11
Refactor Create methods for bool operator values.

Choc13:


src/Computation/Values/And.cs
```diff
@@ -33,17 +34,22 @@ private static IValue ShortCircuit(IValue left, ConstantUnsigned right)
             ? right
             : right.Not().IsZero
                 ? left
-                : new And(left, right);
+                : Create(left, right);
+    }
+
+    private static IValue Create(IValue left, ConstantUnsigned right)
```
Choc13, line 40: The reason for this refactor is so that when I add the associativity changes I can just change this method to be:

```csharp
return left is IConstantValue cl
    ? cl.AsUnsigned().And(right)
    : left is And l
        ? new And(l._left, Create(l._right, right))
        : new And(left, right);
```

This way the recursive part only affects this method alone rather than it constantly calling `ShortCircuit`, it's a minor optimisation, but it also just felt more correct and simpler to comprehend. The `public Create` method just checks if either is a constant and then attempts to short circuit based on the right one definitely being a constant. If that falls through then it calls the `private Create` which is responsible for attempting to merge together constant values by either doing the constant operation if both are constant or pushing the single constant value as far to the right as it can until it either encounters another constant or hits the end of the `Or` train.

src/Computation/Values/And.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Microsoft.Z3;
+﻿using System.IO.Pipes;
```
iscottb122, line 1: Pipes, eh?

tests/Computation.Tests/Values/TestData/UnaryTestData copy.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        foreach (var x in Values())
+            Add(x, new SymbolicUnsigned(x));
+    }
+
+    public static IEnumerable<ConstantUnsigned> Values()
+    {
+        return Enumerable.Range(-8, 24).Select(v => ConstantUnsigned.Create((Bits) 4U, v));
+    }
+}
```
iscottb122, line 21: This is an interesting file name.

src/Computation/Values/And.cs
iscottb122: Logic seems fine.

tests/Computation.Tests/Values/TestData/UnaryTestData copy.cs
iscottb122: Never mind. As discussed.

PR #147, opened 2022-02-13
Fix bug where NOT (and some XOR) are not treated as bitwise.

iscottb122:


tests/Computation.Tests/Values/AndTests.cs
```diff
@@ -14,74 +11,36 @@ public class AndTests
...
-        var symbolic = And.Create(symbolicLeft, symbolicRight).AsBool(Context).Simplify();
+        var result0 = And.Create(left0, right0).AsBool(Context).Simplify();
+        var result1 = And.Create(left1, right1).AsBool(Context).Simplify();
 
-        constant.Should().BeEquivalentTo(symbolic);
-    }
-
-    [Theory]
-    [ClassData(typeof(UnaryTestData))]
-    private void ShouldShortCircuitToZeroWhenEitherArgumentIsZero(
```
Choc13, line 52: It's not clear to me that the above tests cover this. I'm probably missing something, but my understanding is that the above tests assert that things are equal when going through the various symbolic and constant permutations (providing the symbolic actually contains a constant value). 

Whereas these tests are proving that some identity laws are held up. Isn't it still possible to break the implementation of these boolean operators for the identity cases but the above tests pass because the symbolic and constant paths both contain the same error?

Choc13 (approved):

PR #148, opened 2022-02-14
Rename model to constraints.

iscottb122:


src/Computation/Constraints.cs
```diff
@@ -48,13 +48,13 @@ public BigInteger Evaluate(IValue value)
...
         where TContext : IContext, new()
     {
         var context = new TContext();
         var solver = context.CreateSolver();
         solver.Assert(assertions.Select(a => a.AsBool(context)).ToArray());
 
-        return new Model(context, solver);
+        return new Constraints(context, solver);
     }
 }
```
iscottb122, line 60: On my travels I discovered that this should never really have been called `Model` since that would be an interpretation of the constraints rather than just the constraints themselves.

src/Computation/SymbolicProposition.cs
```diff
@@ -6,32 +6,32 @@ namespace Symbolica.Computation;
...
+    public bool CanBeTrue => _constraints.IsSatisfiable(_assertion);
 
     public void Dispose()
     {
-        _model.Dispose();
+        _constraints.Dispose();
     }
 
-    public static IProposition Create(IPersistentSpace space, IValue assertion, IValue[] constraints)
+    public static IProposition Create(IPersistentSpace space, IValue assertion, IValue[] assertions)
```
iscottb122, line 31: This looks weird that we now have `assertion` and `assertions` next to each other but the latter goes away in the next PR so I don't care enough.

src/Computation/IConstraints.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Symbolica.Computation;
 
-internal interface IModel : IDisposable
+internal interface IConstraints : IDisposable
 {
     bool IsSatisfiable(IValue assertion);
```
Choc13, line 9: `AreSatisfiable`?

src/Computation/IConstraints.cs
```diff
@@ -4,7 +4,7 @@
...
     BigInteger Evaluate(IValue value);
```
Choc13, line 10: Maybe I'm understanding this incorrectly, but would it make sense to rename the `IEnumerable<KeyValuePair<string, string>> Evaluate()` method to `GetModel`? Isn't that what Z2 calls it when you ask for a set of values that satisfy the constraints?

src/Computation/IConstraints.cs
iscottb122: It's asking if the assertion passed in is satisfiable within the constraints, so I think it is fine.

src/Computation/IConstraints.cs
iscottb122: Z2 👎 

You might be right but have another look at it on the next PR. It will be clearer then what the various things are. That PR will be quite involved so I was just getting this renaming out of the way because it will be tricky enough to review without all the noise.

Choc13 (approved):

PR #149, opened 2022-02-14
Finally take assertions off of Expression.

iscottb122:


src/Computation/Context.cs
```diff
@@ -7,31 +8,77 @@ internal sealed class Context<TContextHandle> : IContext
...
+        return Create(func);
+    }
+
+    private Solver CreateSolver()
+    {
+        return Create(c => c.MkSolver(c.MkTactic("smt")));
+    }
+
+    private TZ3Object Create<TZ3Object>(Func<Context, TZ3Object> func)
+        where TZ3Object : Z3Object
```
Choc13, line 73: I think we should drop this constraint and just call the generic type parameter `T` here as we don't use the constraint and the method is `private`.

Choc13 (approved):

PR #150, opened 2022-02-14
Stop throwing generic Exceptions from Computation domain.

iscottb122:


iscottb122:
These things (or their predecessors) used to live at the `Application` layer so we weren't bothered about domain exceptions, but after moving them down to `Computation` I forgot to do this.

Choc13 (approved):

PR #164, opened 2022-03-07
Short circuit ArithmeticShiftRight when the value is zero.

Choc13:


src/Computation/Values/ArithmeticShiftRight.cs
```diff
@@ -25,7 +24,8 @@ public static IValue Create(IValue left, IValue right)
         return (left, right) switch
         {
             (IConstantValue l, IConstantValue r) => l.AsSigned().ShiftRight(r.AsUnsigned()),
-            (_, IConstantValue r) when r.AsUnsigned() == BigInteger.Zero => left,
+            (_, IConstantValue r) when r.AsUnsigned().IsZero => left,
+            (IConstantValue l, _) when l.AsUnsigned().IsZero => l,
```
Choc13, line 28: Ah actually, should this one be `l.AsSigned().IsZero` or will the results always be the same regardless? Is negative zero something we need to worry about?

iscottb122: There shouldn't be any such thing as negative zero, and zero should have the same bit representation for unsigned and signed, therefore I think this is fine. I'm also pretty confident the tests will cover this, so if they pass I am happy with it.

iscottb122 (approved):

PR #167, opened 2022-03-09
Ensure bitwise logical operators short circuit in the associative case.

Choc13:


src/Computation/Values/And.cs
```diff
@@ -34,17 +34,18 @@ private static IValue Create(IValue left, ConstantUnsigned right)
         return left switch
         {
             IConstantValue l => l.AsUnsigned().And(right),
-            And l => new And(l._left, Create(l._right, right)),
+            And l => Create(l._left, Create(l._right, right)),
```
Choc13, line 37: This was the problem. If `Create(l._right, right)` hits the `IConstantValue l` case and the result of that `And` on the two unsigned values returns `0` then we should be able to short circuit again, so by calling `Create` we put this new value back through the grinder.

The specific example I saw was with two constants of `1` and `244`.

src/Computation/Values/And.cs
```diff
@@ -34,17 +34,18 @@ private static IValue Create(IValue left, ConstantUnsigned right)
...
     }
 
     public static IValue Create(IValue left, IValue right)
     {
-        return right is IConstantValue r
-            ? ShortCircuit(left, r.AsUnsigned())
-            : left is IConstantValue l
-                ? ShortCircuit(right, l.AsUnsigned())
-                : new And(left, right);
+        return (left, right) switch
```
Choc13, line 44: I also find the style more readable and amenable to any additional cases we want to treat, so I've converted them all. Can revert if you don't agree.

iscottb122 (approved):

PR #169, opened 2022-03-09
Short circuit Multiply when either side is one.

Choc13:


iscottb122 (approved):

src/Computation/Values/Multiply.cs
```diff
@@ -21,8 +21,12 @@ public override BitVecExpr AsBitVector(IContext context)
 
     public static IValue Create(IValue left, IValue right)
     {
-        return left is IConstantValue l && right is IConstantValue r
-            ? l.AsUnsigned().Multiply(r.AsUnsigned())
-            : new Multiply(left, right);
+        return (left, right) switch
+        {
+            (IConstantValue l, IConstantValue r) => l.AsUnsigned().Multiply(r.AsUnsigned()),
+            (IConstantValue l, _) when l.AsUnsigned().IsOne => right,
```
Choc13, line 27: As discussed, switched this to `AsUnsigned().IsOne` because the `Multiply` operator does care about the sign bit when it's doing its thing, it's only the later interpretation of the value that might use a sign bit.

PR #171, opened 2022-03-10
Short circuit Select when consequent and alternative are equal constants.

Choc13:


iscottb122 (approved):

src/Computation/Values/Select.cs
```diff
@@ -47,10 +47,11 @@ public FPExpr AsFloat(IContext context)
...
     {
-        return predicate is IConstantValue p
-            ? p.AsBool()
-                ? trueValue
-                : falseValue
-            : new Select(predicate, trueValue, falseValue);
+        return (predicate, trueValue, falseValue) switch
+        {
+            (IConstantValue p, _, _) => p.AsBool() ? trueValue : falseValue,
+            (_, IConstantValue t, IConstantValue f) when t.AsUnsigned() == f.AsUnsigned() => t,
```
Choc13, line 53: This `==` is wrong. It's comparing the instances of `ConstantUnsigned` which don't implement `IEquatable`, rather than their `BigInteger` values.

iscottb122: Sorry, I should have caught this. I have even been careful to avoid this myself. Just wasn't paying enough attention.

src/Computation/Values/Select.cs
```diff
@@ -47,10 +47,11 @@ public FPExpr AsFloat(IContext context)
...
     {
-        return predicate is IConstantValue p
-            ? p.AsBool()
-                ? trueValue
-                : falseValue
-            : new Select(predicate, trueValue, falseValue);
+        return (predicate, trueValue, falseValue) switch
+        {
+            (IConstantValue p, _, _) => p.AsBool() ? trueValue : falseValue,
+            (_, IConstantValue t, IConstantValue f) when (BigInteger) t.AsUnsigned() == (BigInteger) f.AsUnsigned() => t,
```
iscottb122, line 53: I think we should do this as

```
(_, IConstantValue t, IConstantValue f) when t.AsUnsigned().Equal(f.AsUnsigned()) => t,
```

using the `Equal` method (not `Equals`) on `ConstantUnsigned` to produce a `ConstantBool` which is then implicitly casted to `bool`. This whole design is maybe not perfect but assuming we are not going to rewrite it all at the moment it is consistent with how everything else works.

iscottb122 (approved):

PR #175, opened 2022-03-11
Always go via constraints when dealing with contexts.

iscottb122:


iscottb122:
This provides the `IConstraints` abstraction I need, but also fixes a rogue issue that has been hiding, where a constant value could be returned which violates symbol assertions.

src/Computation/Constraints.cs
```diff
@@ -33,23 +38,29 @@ public bool IsSatisfiable(IValue assertion)
         };
     }
 
-    public BigInteger Evaluate(IValue value)
+    public BigInteger GetConstant(IValue value)
+    {
+        var constant = value.AsConstant(_context);
+
+        return constant == GetValue(value)
```
Choc13, line 45: Just for my curiosity can you explain this check?

iscottb122: Yes. Previously `Expression` directly tried to simplify whatever value it held, and then got the numeric value from that. This would never need to touch a solver, but it was incorrect because it would ignore any symbol-level assertions. The correct logic is to still do this to get the one and only value, but also use a solver get a model, evaluate the value for this model, and check that this example value is the same as the one and only possible value.

Choc13 (approved):

PR #185, opened 2022-03-18
Wrap all Z3Objects in a using statement.

Choc13:


src/Computation/Assertions.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Z3;
+
+namespace Symbolica.Computation;
+
+internal sealed class Assertions : IEnumerable<BoolExpr>, IDisposable
```
Choc13, line 9: I introduced this so that we have a disposable collections of assertions. It has the upside of making the `IContext.Assert` api safer and makes the ownership / disposal of the assertions clearer (i.e. whoever constructs this object should wrap it in a `using`).

iscottb122 (approved):

PR #186, opened 2022-03-18
Generate the Exprs from IValues inside Context.

iscottb122:


src/Computation/PooledContext.cs
```diff
@@ -64,9 +64,11 @@ public BigInteger GetSingleValue(BitVecExpr variable)
             : throw new IrreducibleSymbolicExpressionException();
     }
 
-    public BigInteger GetExampleValue(BitVecExpr variable)
+    public BigInteger GetExampleValue(IValue variable)
     {
-        return ((BitVecNum) CreateModel().Eval(variable, true)).BigInteger;
+        var expr = variable.AsBitVector(this);
+
+        return ((BitVecNum) CreateModel().Eval(expr, true)).BigInteger;
```
iscottb122, line 71: This was what I was always trying to avoid. If `expr` were to be inlined this becomes wrong since any implicit symbol assertions it makes are not applied to the model. I think my cure of enforcing that the method take the `Expr` was worse than the disease but I still don't like this. We are about to add a `using` so inlining does become less likely, but still seems dodgy.

src/Computation/PooledContext.cs
```diff
@@ -27,20 +27,20 @@ public void Dispose()
         Contexts.Add(_context);
     }
 
-    public void Assert(IEnumerable<BoolExpr> assertions)
+    public void Assert(IEnumerable<IValue> assertions)
     {
-        _solver.Add(assertions);
+        _solver.Add(assertions.Select(a => a.AsBool(this)));
```
Choc13, line 32: So we do away with my `Assertions` type now and just do the Dispose loop here?

Choc13 (approved):

src/Computation/PooledContext.cs
Choc13: Impérative languages for you 🤷‍♂️ At least we've localised to this method. I do think a comment here to our future selves would be a good idea though.

src/Computation/PooledContext.cs
iscottb122: Yeah, either use the `Add` method on the solver one assertion at a time in a loop or do them all at once then dispose them all at once. Probably the former since I'm pretty sure there is nothing more efficient going on in there.

PR #187, opened 2022-03-19
It's really just a solver that exposes its Context!

iscottb122:


src/Computation/Values/FloatPower.cs
```diff
@@ -17,30 +17,28 @@ private FloatPower(IValue left, IValue right)
...
+            : AsReal(solver, _left);
+
+        using var right = _right is IRealValue r
+            ? r.AsReal(solver)
+            : AsReal(solver, _right);
+
+        return (RealExpr) solver.Context.MkPower(left, right);
+    }
+
+    private static RealExpr AsReal(ISolver solver, IValue value)
```
Choc13, line 38: I think this is a good example of where a local function is better tbh.

src/Computation/Values/FloatPower.cs
```diff
@@ -17,30 +17,28 @@ private FloatPower(IValue left, IValue right)
...
-                })
-            };
-
-        return context.CreateExpr(c =>
-        {
-            using var left = ToArithExpr(_left);
-            using var right = ToArithExpr(_right);
-            return (RealExpr) c.MkPower(left, right);
-        });
+        using var left = _left is IRealValue l
```
Choc13, line 27: I also think this was better as a local function too in order to avoid this duplication.

src/Computation/Values/FloatUnordered.cs
```diff
@@ -14,20 +14,15 @@ private FloatUnordered(IValue left, IValue right)
...
-            }
+        using var left = _left.AsFloat(solver);
+        using var leftIsNaN = solver.Context.MkFPIsNaN(left);
 
-            using var left = MkFPIsNaN(_left);
-            using var right = MkFPIsNaN(_right);
-            return c.MkOr(left, right);
-        });
+        using var right = _right.AsFloat(solver);
+        using var rightIsNaN = solver.Context.MkFPIsNaN(right);
```
Choc13, line 23: Think the local function was nicer again here to dedupe.

Choc13 (approved):

PR #188, opened 2022-03-19
Use solver for single value instead of attempting to extract numeric.

iscottb122:


iscottb122:
There was a latent bug in here where values that could not be simplified to a single value but only have one possible interpretation in a model would throw. We never hit it since this operation is pretty rare anyway, but might as well make it more general purpose and sensible anyway.

iscottb122:
I imagine we might want to change this to some kind of `TryGetSingleValue` (internally at least) as you mentioned before. I think both of our implementations of the read/write stuff use versions of this in various places. I will leave it for now since there is another related idea I want to propose first.

src/Computation/PooledSolver.cs
```diff
@@ -58,23 +60,18 @@ public bool IsSatisfiable(IValue assertion)
...
-        using var simplified = bitVector.Simplify();
+        var constant = GetExampleValue(value);
 
-        var value = simplified.IsNumeral
-            ? ((BitVecNum) simplified).BigInteger
-            : throw new IrreducibleSymbolicExpressionException();
-
-        return value == GetExampleValue(variable)
-            ? value
+        return IsSatisfiable(NotEqual.Create(value, ConstantUnsigned.Create(value.Size, constant)))
```
Choc13, line 67: Shouldn't these branches be the other way around? I.e if this is false because there aren't any other constants it can be then return the constant.

iscottb122: This is why one should always run tests.

Choc13 (approved):

PR #189, opened 2022-03-19
Move the NaN special case check into float itself.

iscottb122:


iscottb122:
This was the facepalm one I was talking about recently. I can't believe how long I have been carrying this stupid special case thing around when I could just put it in the float class.

Choc13 (approved):

PR #193, opened 2022-03-20
Add a solver which will only go symbolic when necessary.

iscottb122:


src/Computation/LazySolver.cs
```diff
@@ -0,0 +1,73 @@
...
+    public Context Context => GetSolver().Context;
+
+    public void Dispose()
+    {
+        _solver?.Dispose();
+    }
+
+    public void Assert(IEnumerable<IValue> assertions)
+    {
+        _actions.Add(s => s.Assert(assertions));
```
Choc13, line 29: What about assertions that are added after the solver is created?

src/Computation/LazySolver.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Numerics;
+using Microsoft.Z3;
+using Symbolica.Expression;
+
+namespace Symbolica.Computation;
+
+internal sealed class LazySolver : ISolver
+{
+    private readonly IList<Action<ISolver>> _actions;
+    private ISolver? _solver;
```
Choc13, line 12: Can we not use `Lazy` here? Or are se worried about serialisation? Although I didn't think these were ever serialised. 

I guess right now we're also threadsafe because of the fact the `PooledSolver` is and that we only use a `Solver` on a particular branch / state before disposing of it anyway.

src/Computation/Proposition.cs
```diff
@@ -3,14 +3,14 @@
 
 namespace Symbolica.Computation;
 
-internal sealed class SymbolicProposition : IProposition
```
Choc13, line 6: So we're just happy to create these unconditionally of whether it's symbolic or not because something else will take care of ignoring the constant ones?

src/Computation/LazySolver.cs
iscottb122: Should be fine and follow the same path as any done before the solver is created. Unless I have messed up the implementation of course.

Choc13: Ah I see now. Everything goes through `GetSolver()` to get the solver and so the actions are always applied before the solver is returned. I thought it was just creation time laziness. 👍

src/Computation/LazySolver.cs
iscottb122: Perhaps this will be clearer once the above misunderstanding is resolved. To be clear thought, `PooledSolver` is definitely not meant to be thread-safe (except for creating it using the pool of `Context`s).

Choc13 (approved):

src/Computation/Proposition.cs
iscottb122: Yeah, we should discuss this more generally. I think it is useful in both of our approaches to reading/writing etc, but I could be wrong.

PR #194, opened 2022-03-21
Short circuit Select when consequent and alternative are equivalent expressions.

Choc13:


src/Computation/Values/Select.cs
```diff
@@ -50,6 +50,7 @@ public static IValue Create(IValue predicate, IValue trueValue, IValue falseValu
         {
             (IConstantValue p, _, _) => p.AsBool() ? trueValue : falseValue,
             (_, IConstantValue t, IConstantValue f) when t.AsUnsigned().Equal(f.AsUnsigned()) => t,
+            _ when trueValue.Equals(falseValue) => trueValue,
```
Choc13, line 53: Related to your other comment about constant equality, this is a case where we don't care about the whether the constant is signed or unsigned etc. So for now I've left the constant case in which then explicitly does the comparison as unsigned.

PR #196, opened 2022-03-21
Use the new zero everywhere.

iscottb122:


src/Computation/Values/Constants/ConstantUnsigned.cs
```diff
@@ -161,18 +160,18 @@ public ConstantUnsigned Xor(ConstantUnsigned value)
         return value._value;
     }
 
+    public static ConstantUnsigned CreateZero(Bits size)
+    {
+        return new ConstantUnsigned(size, BigInteger.Zero);
+    }
+
```
iscottb122, line 167: This is obviously a nonsense PR but zero doesn't have to go through the normalization (since it is zero). This only really makes a difference for very large ones but they do occur on occasion.

Choc13 (approved):

PR #198, opened 2022-03-29
Write whole block in MemorySet rather than each byte individually.

Choc13:


src/Representation/Functions/MemorySet.cs
```diff
@@ -36,13 +36,15 @@ public SetMemory(IExpression destination, IExpression value)
...
-        public void Invoke(IState state, BigInteger value)
+        public void Invoke(IState state, BigInteger bytes)
         {
-            foreach (var offset in Enumerable.Range(0, (int) value))
-            {
-                var address = _destination.Add(state.Space.CreateConstant(_destination.Size, offset));
-                state.Memory.Write(address, _value);
-            }
+            var size = ((Bytes) (uint) bytes).ToBits();
+            var buffer = state.Space.CreateConstant(size, BigInteger.Zero);
```
Choc13, line 42: Do you think we should introduce `CreateZero` in the same way we have `ConstantUnsigned.CreateZero` in `Computation`?

iscottb122 (approved):

src/Representation/Functions/MemorySet.cs
iscottb122: Leave it for now.

Choc13: Happy to do it on a completely separate PR.

PR #208, opened 2022-04-22
Use bytes for types instead of bits.

iscottb122:


src/Representation/Types/StructType.cs
```diff
@@ -25,23 +26,26 @@ public IType GetType(ISpace space, IExpression index)
 
     public IExpression GetOffsetBits(ISpace space, IExpression index)
     {
-        return space.CreateConstant(space.PointerSize, (uint) _offsets[(int) index.GetSingleValue(space)]);
+        return space.CreateConstant(space.PointerSize, (uint) _offsets[(int) index.GetSingleValue(space)].ToBits());
```
Choc13, line 29: Could this not be `GetOffsetBytes().ToBits()`

Choc13 (approved):

src/Representation/Types/StructType.cs
iscottb122: It's an expression already so this is the only layer I can do it at. Could do a multiply but probably worse that way.

PR #214, opened 2022-05-04
Equatable IExpression

Choc13:


Choc13:
The first commit is me fixing up equality for those Value records that contain reference types and also making sure the default Equals for all constant Values goes through the `ConstantUnsigned` path rather than just doing structural equality on itself.

iscottb122 (approved):

src/Computation/Values/Constants/ConstantUnsigned.cs
```diff
@@ -55,7 +55,12 @@ public ConstantDouble AsDouble()
...
     public override bool Equals(IValue? other)
     {
-        return other is IConstantValue v && Equal(v.AsUnsigned());
+        return Equals(other as IConstantValue);
+    }
+
+    public bool Equals(IConstantValue? other)
+    {
+        return other is not null && Equal(other.AsUnsigned());
     }
```
iscottb122, line 65: I assume this gets `GetHashCode` automatically as a `record`.

Choc13: Yeah and in this case we do want the default. It's all the other Constants that call through to this one which is why I changed their behaviours and left this one. Same for Equals.

PR #217, opened 2023-02-05
Add quick-start instructions to Readme

richthornton:
Hello @iscottb122! I was wanting to play around with what you legends have made, and the instructions to get started weren't too clear, so @Choc13 gave me some pointers. I thought it would be useful to document these just in case anyone else wants to try it out (or I'm missing an easier way).

As an aside, the reason I was excited to try it out was because I think it becomes even more important with AI powered code generation. As we all know, they can be amazing, but also confidently wrong. Here's a video of how nice it is having them both together:

https://user-images.githubusercontent.com/7372040/216797544-ccfa095b-90b0-4c2f-9fef-5a3ad876a5cc.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,43 @@
 # Symbolica
 
+Run your code forall inputs
+
+## What is Symbolica?
+Symbolica is a symbolic code executor hosted as a 'CI-like' service. We run your code, but instead of feeding it concrete inputs like 1 or "Hello World", we feed it symbols (variables) that represent all of the possible inputs to your program. This allows us to explore all the code paths at the same time, which allows us to do all kinds of interesting things, such as:
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,43 @@
...
+
+For example:
+```sh
+export SYMBOLICA_BUILD_IMAGE=symbolica/build:0.2.0-master0183
+export SYMBOLICA_TRANSLATE_IMAGE=symbolica/translate:0.2.0-master0183
+```
+
+### Run the example
+```sh
+dotnet run --project src/Application './example'
```
Choc13, line 37: I think your example is the same as the test in `tests/Application.IntegrationTests/pass/sign`. So we could probably just add the comments you added to the source file there and reference that one instead.

